PR #2, opened 2018-11-13
Add AIOHTTP as alternative server framework

matthewrmshin:
AIOHTTP is also a good candidate to consider for web server. Its demos have examples on WebSocket and GraphQL usages.


kinow:
Looks interesting! Quite similar to Tornado, but with a few things from Flask too. And a few benchmarks from their website appear to provide good reason for including it in the tests. +1

PR #9, opened 2018-11-28
document benefits and motivation behind the new architecture

oliver-sanders:
The six points from the June meeting.

### DIFF
--- a/docs/cylc-8-roadmap.md
+++ b/docs/cylc-8-roadmap.md
@@ -13,7 +13,8 @@ meetings, but not in a way that will hurt you._
 We need something very much like the [Jupyter
 Hub](https://jupyterhub.readthedocs.io/en/stable/) Architecture: a privileged
 hub that spawns a reverse proxy server and sub-services for suite discovery etc.,
-and spawns suite daemons (c.f. Jupyter notebooks) as the user.
+and spawns UI servers (c.f. Jupyter notebooks) and potentially suite daemons as
+the user.
 
 ### Core Principles & Motivation For The New Architecture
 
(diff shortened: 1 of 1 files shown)


docs/cylc-8-roadmap.md
```diff
@@ -13,7 +13,23 @@ meetings, but not in a way that will hurt you._
 We need something very much like the [Jupyter
 Hub](https://jupyterhub.readthedocs.io/en/stable/) Architecture: a privileged
 hub that spawns a reverse proxy server and sub-services for suite discovery etc.,
-and spawns suite daemons (c.f. Jupyter notebooks) as the user.
+and spawns UI servers (c.f. Jupyter notebooks) and potentially suite daemons as
```
oliver-sanders: After some very quick tinkering with JupyterHub I think our "UI Server" fits onto the place of the Jupyter Notebook:

When you login to JupyterHub it spawns (if required) and redirects you to your notebook. For us we would want the user to be taken to the GUI (provided by the UI server) so they can see a list of running suites etc.

Which leaves us with the question of what spawns the suites:

1. The UI server, it is running as `$USER` so this should be no issue *BUT* if done this way the UI server must either:
    1. Inform the hub
    2. Spin up its own proxy server and manage its own suites
2. The hub, more awkward as the hub is privileged but better for centralisation.

I'm guessing 1.1 is likely to be the preferred option.

docs/cylc-8-roadmap.md
```diff
@@ -13,7 +13,23 @@ meetings, but not in a way that will hurt you._
...
+the user.
+
+### Core Principles & Motivation For The New Architecture
```
sadielbartholomew: The first section under 'Architectural Challenges' in the ``gui-replacement-options.md`` file discusses this topic - can we move one of these two sections/lists so they are in the same place, or even collate them together, to keep the material organised? @hjoliver I personally feel the 'Architectural Challenges' would be better placed in this file, as a preference for movement direction.

docs/cylc-8-roadmap.md
hjoliver: (I'll put this detailed description in the roadmap doc).

docs/cylc-8-roadmap.md
hjoliver: @sadielbartholomew - sounds sensible.  I'll merge this PR and then tweak along the lines you suggest.

PR #13, opened 2018-12-14
Report assignment to docs conversion

sadielbartholomew:
I started tackling cylc/cylc#2651 (the basic conversion) yesterday afternoon, so to ensure nobody else begins this work please add my name by it in the task list. It's not critical work, but I think I might as well get it done, so it is not an interruption.

So far, I have got the Sphinx project set-up (& created a simple but very nice slightly-customised theme which I hope we will all like & agree to go with) & lifted & shifted the CUG & SDG. Once I get the LaTeX translated to reStructuredText I will open a WIP PR (estimate: early next week) & outline input needed by the team RE what to include, structure, theme etc.


hjoliver:
Thanks @sadielbartholomew, good news.

PR #15, opened 2019-01-07
Update proposal-rose-suite-run.md

matthewrmshin:
Fix typos + new minor points to consider on job cluster support.


matthewrmshin:
@hjoliver I don't have permission to assign reviewer, but please review.

hjoliver:
> @hjoliver I don't have permission to assign reviewer, but please review.

I think I've just fixed that.

PR #20, opened 2019-03-15
Suggest coding styles for the meeting agenda of Mar/2019

kinow:
Quick pull request with a suggestion for coding conventions for our repositories. Our Python code is being rewritten in some packages, so this is a good time to establish code conventions to be followed.

These conventions could also be re-used in the Cylc UI Server and other Python projects.

There are already a few docstrings, but some in the google format, some in the old PEP 257, some in the docstring supported by PyCharm (_cough, cough_, sorry).

I think ideally we need one that can be used for documentation, but also by syntax highlighters and IDE's - the more tools supported, the merrier IMO.

Cheers
Bruno

### DIFF
--- a/docs/vc-march-2019-agenda.md
+++ b/docs/vc-march-2019-agenda.md
@@ -90,6 +90,8 @@
   - Do we need to choose coding styles?
       * Python Docstring Conventions - would be nice to have one that works with most tools (i.e. sphinx, vim syntax hightlight, IDE, napoleon, etc)
       * A `CONTRIBUTING.md` with guidelines for people creating PR for the first time (the 2 approvers rule is nice, but not well known until you send a contribution), also explaining that test coverage is desired when possible, as well as documentation update, etc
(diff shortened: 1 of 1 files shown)


kinow:
This silly GitHub UI... I thought the branch would be created in my repo. So will delete the `kinow-patch-1` branch once closed/merged.

matthewrmshin:
Maybe worth filling in the full set of community profile e.g. https://github.com/cylc/cylc-admin/community as well.

And should we use something like [Black](https://github.com/ambv/black)?

kinow:
Good idea @matthewrmshin , will update the branch in the next minutes. Thanks!

hjoliver:
Here's the cylc/cylc one: https://github.com/cylc/cylc/community

We have CONTRIBUTING.md now, only code of conduct is missing (not sure we need issue and PR templates? ... maybe if we start getting much from totally new people).

But I guess we need the same for the other new projects.

hjoliver:
(Branch deleted)

kinow:
Thanks @hjoliver !

PR #29, opened 2019-06-11
Added review comments from suites guild.

wxtim:


hjoliver:
Yes, good work :bouquet:

wxtim:
> Surprised by the number of people answering that never heard about some utilities. 
We have a huge range of use cases - I came into the team from post processing, so I've seen a lot of jinja2 :cry:, but had no idea about some of the other stuff.

> no mention of the docs 
I noticed this too - again possibly a f(n) of a large user community.

@hjoliver  - I've added it to the index.

oliver-sanders:
> Surprised by the number of people answering that never heard about some utilities.

We're not at all surprised!

Particularly on the Cylc side (due to Cylc being hidden away by the Rose wrapper). When I wrote the tutorials in the Rose documentation I made a special effort to get people using commands like:

* `cylc graph`
* `cylc graph -n`
* `cylc get-config --sparse`

These commands are unknown to many users.

matthewrmshin:
Hi @kinow On *gpanel*. I believe users still want an always-on widget in their desktop tray or notification area.

wxtim:
> Hi @kinow On _gpanel_. I believe users still want an always-on widget in their desktop tray or notification area.

Do they want this? Or do they want something else? Would some users prefer a pop up notification if something fails/finishes, for example?

hjoliver:
What is the official "always-on widget" paradigm in recent versions of Gnome 3 desktop? Is there one?

oliver-sanders:
https://www.omgubuntu.co.uk/2017/09/will-you-miss-gnome-legacy-tray

oliver-sanders:
There is a Gnome page advising app developers on what to do with application status:

https://wiki.gnome.org/Initiatives/StatusIconMigration/Guidelines

I think the main use case for `gpanel` is to be kept up-to-date with job failures and anything which might require user intervention. For this Gnome notifications might be a viable option.

hjoliver:
Sorry @wxtim I just noticed the location of you new file. I won't get published from there, to GitHub Pages.  From repo settings:
Screenshot from 2019-06-12 22-19-52

Suggest moving `research/` into `docs/research/` and updating the index link accordingly (c.f. `docs/meetings/` content).

hjoliver:
That Gnome "status icon" guidelines document is hating on gpanel big time.  Looks like it'll have to be notifications, or a permanent small main window.

To be honest I stopped using gpanel several years ago, when I started using gnome 3.

matthewrmshin:
> To be honest I stopped using gpanel several years ago, when I started using gnome 3.

@hjoliver We have only moved from RHEL6 (Gnome 2) to RHEL7 (Gnome 3) some weeks ago!

wxtim:
@hjoliver 
> Suggest moving research/ into docs/research/ and updating the index link accordingly (c.f. docs/meetings/ content).

This I have now done.

wxtim:
I'm almost tempted to add some points from this conversation to the document. That way lies recursive madness.

sadielbartholomew:
> I'm almost tempted to add some points from this conversation to the document. That way lies recursive madness.

In seriousness, though, please don't do this, as (save for careful adding excessive annotations) it will mix up our own thoughts with user feedback. Going forward, it is important we can easily distinguish between what has been said by *users* (which is the crucial part) & by *us* developers (especially if it includes our assumptions or guesses about what users might think).

PR #30, opened 2019-06-12
gcylc research: report of gcylc.rc usage at MO

oliver-sanders:
Here is how Met Office Cylc users are using the current (soon to be extinct) `gcylc.rc` configuration.

Will add to the index once #29 is merged.


research/MO-gcylc-rc-report.md
```diff
@@ -0,0 +1,117 @@
...
+  * ~700 users in total.
+  * ~150-200 simultaneous users at any one time.
+  * 76 users have gcylc.rc files.
+  * There isn't much customisation of `gcylc` at present.
+  * A lot of people are hard-coding the default settings.
+
+* Views:
+  * The most changed setting is `initial views` people like to customise this.
+  * People who work with large suites often disable the graph setting.
+  * There isn't much love for the `dot` view.
```
hjoliver: Aww come on people, show me the love!

hjoliver:
This is useful, but probably doesn't capture (the majority of?) users who aren't aware of this configurability and thus "don't know what they're missing"?

oliver-sanders:
Yup

oliver-sanders:
FYI I auto-generated this report with the following script:

```bash
#!/bin/bash


set -eu
shopt -s extglob


extract () {
    for gcylc in /home/*/*/.cylc/gcylc.rc; do
        cp "$gcylc" "$(echo "$gcylc" | cut -d '/' -f 4)"
    done
}


search () {
    local pattern="$1"
    for file in !(gcylc); do
        if grep '^[^#]' "$file" | grep -q "$pattern"; then
            echo $file
        fi
    done
}


get_values () {
    local pattern="$1"
    for file in !(gcylc); do
        if grep '^[^#]' "$file" | grep -q "$pattern"; then
            ret="$(sed -rn "s/${pattern}\s+?=\s+?(.*)/\1/p" "$file" \
                   | sed -r 's/#.*$//; s/\s+$//')"
            if grep -q ',' <<< "$ret" && [[ "$pattern" != "window size" ]]; then
                # list options as set
                echo $ret
                # break down options into components
                ret=$(sed 's/ //g' <<< "$ret")
                local IFS=','
                local retb
                read -ra retb <<< "$ret"
                for item in "${retb[@]}"; do
                    if [[ "$item" != '' ]]; then
                        echo "$item"
                    fi
                done
            elif [[ "$ret" != '' ]]; then
                echo "$ret"
            fi
        fi
    done
}


KEYS=(\
    'dot icon size' \
    'initial side-by-side views' \
    'initial views' \
    'maximum update interval' \
    'sort by definitiion order' \
    'sort column' \
    'sort column ascending' \
    'sub-graphs on' \
    'task filter highlight color' \
    'task states to filter out' \
    'transpose dot' \
    'transpose graph' \
    'ungrouped views' \
    'use theme' \
    'window size' \
    '\[themes\]' \
)

VALUES=(\
    'dot icon size' \
    'initial side-by-side views' \
    'initial views' \
    'maximum update interval' \
    'sort by definition order' \
    'sort column' \
    'sort column ascending' \
    'sub-graphs on' \
    'task filter highlight color' \
    'task states to filter out' \
    'transpose dot' \
    'transpose graph' \
    'ungrouped views' \
    'use theme' \
    'window size' \
)


report_all() {
    for pattern in "${KEYS[@]}"; do
        number=$(search "$pattern" | wc | awk '{ printf "%02d", $1 }')
        echo "$number $pattern"
    done
}


report_values() {
    for pattern in "${VALUES[@]}"; do
        echo "# $pattern"
        get_values "$pattern" | sort | uniq -c | sort -r
    done
}


report () {
    report_all | sort -r
    report_values
}


COMMAND="$1"; shift
"$COMMAND" "$@"
```

oliver-sanders:
Note: #29 accidentally added two copies of the same file so removing the duplicate here.

sadielbartholomew:
> FYI I auto-generated this report with the following script

Nice!

When did you run the script to capture the information in``MO-gcylc-rc-report.md``? I think it is important we note down the time point for which this data is a snapshot.

It would be nice to make a similar script to scrape for other user-defined visualisation aspects, e.g:

* ``gscan.rc``;
* the ``[visualization]`` section of user's ``suite.rc`` files, if set.

docs/research/MO-gcylc-rc-report.md
```diff
@@ -0,0 +1,117 @@
...
+### `use theme`
+* 7 `default`
+* 5 `solid`
+* 3 `high-contrast`
+* 3 `color-blind`
+* 2 `SMS`
+* 2 `PinkRun`
+* 1 `multi`
+* 1 `custom`
+* 1 `adam`
```
sadielbartholomew: Maybe we could group the non-standard themes (set by only user each) into some umbrella category such as 'user's custom individual theme'? Clearly Adam & Anette (& some other users not naming their themes after themselves) had their own visualisation preferences :smile:!

oliver-sanders: Cylc provides:
* 7 `default`
* 5 `solid`
* 3 `high-contrast`
* 3 `color-blind`

The rest are custom.

sadielbartholomew: The point I was trying to make (but didn't convey at all, sorry) is that it doesn't matter what those themes are called, what is important is that each user has defined their own. So a group to collate all users defining their own theme (i.e. the sum of the totals for all themes not on your list above) would seem more useful to me.

Saying that, there appear to be some user-defined themes e.g. ``PinkRun`` that multiple users have applied. Any idea how that has happened? Can users share themes? (I doubt it is just a coincidental choice of the same name).

oliver-sanders: https://github.com/cylc/cylc-flow/blob/7.8.3/etc/gcylc.rc.eg#L19

sadielbartholomew: Oh dear :joy: Perhaps those poor souls forgot they tried that out & think the default theme uses magenta?

docs/research/MO-gcylc-rc-report.md
```diff
@@ -0,0 +1,117 @@
...
+  * ~700 users in total.
+  * ~150-200 simultaneous users at any one time.
+  * 76 users have gcylc.rc files.
+  * There isn't much customisation of `gcylc` at present.
+  * A lot of people are hard-coding the default settings.
+
+* Views:
+  * The most changed setting is `initial views` people like to customise this.
+  * People who work with large suites often disable the graph setting.
+  * There isn't much love for the `dot` view.
```
wxtim: I like the flowers, I like the daffodils, I like the dot view!

PR #31, opened 2019-06-19
Getting Started with Cylc 8

wxtim:
Uploaded notes on running a dev version of Cylc 8 in a conda environment.


hjoliver:
Thanks @wxtim that's really helpful.  And it looks pretty straightforward too, which is nice!

PR #32, opened 2019-06-20
Update HOWTO cylc-8

matthewrmshin:
Address few issues after I have tried the instructions from scratch.

### DIFF
--- a/docs/howto/environment.md
+++ b/docs/howto/environment.md
@@ -2,11 +2,11 @@
 
 1. Create a clean conda install:
   ```bash
-  conda env create -n my-example-env python=3.7
+  conda create -n my-example-env python=3.7
   conda activate my-example-env
   ```
 
-2. Ensure that you install node-js using conda first*. If you allow pip to
+2. Ensure that you install `nodejs` using conda first*. If you allow pip to
   install it as a dependency of cylc & cylc-uiserver it you will end up with
   a very old version.
   ```bash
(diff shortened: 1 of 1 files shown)


docs/howto/environment.md
```diff
@@ -2,11 +2,11 @@
 
 1. Create a clean conda install:
   ```bash
-  conda env create -n my-example-env python=3.7
+  conda create -n my-example-env python=3.7
   conda activate my-example-env
   ```
 
-2. Ensure that you install node-js using conda first*. If you allow pip to
+2. Ensure that you install `node-js` using conda first*. If you allow pip to
```
wxtim: `nodejs` or node-js. 
`conda install node-js` will fail.

docs/howto/environment.md
```diff
@@ -15,35 +15,41 @@
...
-4. Install npm in cylc-ui client.
+  where `${DEVDIR}/cylc-flow` and `${DEVDIR}/cylc-uiserver` are the locations
+  of your local (git) worktrees of the two projects.
+
+4. Install [cylc/cylc-ui](https://github.com/cylc/cylc-ui/).
   ```bash
   cd $DEVDIR/cylc-ui
-  npm install
+  npm install --prefix="${DEVDIR}/cylc-ui"
   npm install -g configurable-http-proxy
```
matthewrmshin: @kinow Do you know why we need to install `configurable-http-proxy` separately?

docs/howto/environment.md
```diff
@@ -15,35 +15,41 @@
...
 
-8. Run the UI
+7. Run UI.
   ```bash
-  npm run build:watch
+  conda activate my-example-env
+  npm run --prefix="${DEVDIR}/cylc-ui" build:watch
   ```
 
 * Open a browser and navigate to `localhost:8000`. You will be asked to log in
```
wxtim: Not really part of main req, but why not?
Suggest modify to 
[localhost:8000](https://localhost:8000)

docs/howto/environment.md
```diff
@@ -15,35 +15,41 @@
...
   For comparison devs have used node at 10.13.x and 12.x.x.
 
-3. Install development copies of cylc and cylc-uiserver:
+3. Install
+   [cylc/cylc-flow](https://github.com/cylc/cylc-flow/) and
+   [cylc/cylc-uiserver](https://github.com/cylc/cylc-uiserver/):
   ```bash
-  pip install -e $DEVDIR/cylc
-  pip install -e $DEVDIR/cylc-uiserver
+  pip install -e "${DEVDIR}/cylc-flow" -e "${DEVDIR}/cylc-uiserver"
```
kinow: That will change soon. I think `master` branch of `cylc-uiserver` does not rely on `cylc-flow` being installed.

The branch of David that uses GraphQL needs `cylc-flow`, but not the alpha versions. It is installing his fork of cylc-flow with the related GraphQL backend work.

But I think it might still be interesting to install cylc-flow editable for development I think - though the developer may need to edit `setup.py` for the version range

docs/howto/environment.md
kinow: I saw this node-js, and I considered it to be the project name (which can be node, node-js, node.js, NodeJS, etc). For the conda package I think it's `nodejs` - even though there is a `node` package in conda too. Just to confuse people

docs/howto/environment.md
kinow: Probably `https` won't be enabled in development... though that's possible too. Maybe `http://localhost:8000`?

docs/howto/environment.md
matthewrmshin: Changed to `nodejs` for clarity of context with Conda here.

docs/howto/environment.md
matthewrmshin: Modified to an unwrapped `http://localhost:8000/` string.

docs/howto/environment.md
matthewrmshin: Noted. We should update when that happens.

docs/howto/environment.md
kinow: If you install `jupyterhub` via `pip`, then you [need to install the proxy separately](https://jupyterhub.readthedocs.io/en/stable/quickstart.html#prerequisites).

You can install `jupyterhub` via Conda, and that installs the `configurable-http-proxy` as well (it might be useful for our packaging in Cylc now that I think about it...)

Perhaps using conda could be simpler, though I don't know if we have something like editable installations with Conda.

kinow:
2 approvals, but there's one question and I don't know if @matthewrmshin will add to change anything for `configurable-http-proxy`. If not, feel free to hit the merge button @matthewrmshin or @wxtim :grimacing:

matthewrmshin:
OK, `conda install jupyterhub` will end up with a very old version of `nodejs`. So, the current instruction is good.

PR #37, opened 2019-07-09
Exeter meeting notes.

hjoliver:
Stream of consciousness, transcribed quickly from hand-written notes.

Needs organising later.


docs/meetings/exeter-jun-2019.md
```diff
@@ -0,0 +1,254 @@
+# Exeter Meeting Notes
+
+June 24-28
+Present HO, DM, MS, OS, SB, BK
+
+- ZenHub - use instead of Trello, as it provides a powerful multi-project view
+  of GitHub.
```
matthewrmshin: I guess we should start using this. Should discuss in the upcoming VC?

docs/meetings/exeter-jun-2019.md
```diff
@@ -0,0 +1,254 @@
...
+  right network routing.
+
+- JupyterLab is a different/better notebook interface. Has a nice user-movable
+  tiling tab feature that we should try to use. 
+
+- Most tree view info should be retrieved on request only (e.g. Job ID etc. in
+  the current tree view, not needed unless the final level is expanded in the
+  new UI.
+
+- "rose suite-run" - need plugin-based file-installation, e.g. via fcm keywords
```
matthewrmshin: Not sure what this means? I guess you are talking about supporting 3rd party file install schemes? The current functionality in Rose is already modular and plugin friendly. With the packaging change, it should become very easy to install plugins.

docs/meetings/exeter-jun-2019.md
```diff
@@ -0,0 +1,254 @@
...
+
+- rose-suite.conf should be an importable Python file.
```
matthewrmshin: ?

docs/meetings/exeter-jun-2019.md
```diff
@@ -0,0 +1,254 @@
...
+- UIS has to see source dir
+  - need to see diffs of src vs run-dir
+  - MS: could use git (whole suite, minus work tree etc.); tie commits to runs,
+    restarts, reloads.
+
+- Cylc 9:
+  Python API gives modularity
+  parameterizing sub-graphs 
+  spawn-on-demand: avoids need for multi-dimensional cycling (just follow the
+   graph, whether it is generated by cycles or parameters)
```
matthewrmshin: Add (beyond Cylc 8):
* Housekeep - port functionality of `rose_prune` to Cylc, and improve.
* Run a sub-graph as a single job.

docs/meetings/exeter-jun-2019.md
```diff
@@ -0,0 +1,254 @@
...
+
+- Agreed new default installation dir structure:
```
matthewrmshin: Proposed rather than agreed.

docs/meetings/exeter-jun-2019.md
```diff
@@ -0,0 +1,254 @@
...
+```
+    src ---> src/run1
+                /run2
+                /run3
+```
```
matthewrmshin: What about allowing users to parameterize run labels, with the above be the recommended default? Users may want to name their runs with better names.

docs/meetings/exeter-jun-2019.md
```diff
@@ -0,0 +1,254 @@
...
+- UI form for Jinja2 params
+- Note need static graph vis too, in Cylc 8 (tell MR)
+
+- TODO: can we "query the API" over GraphQL (UIS-WFS) so to allow multiple
+  cylc-flow versions on the back end?
+
+- OS agrees with DS PRs in large part; GraphQL schema looks good.
+  - not convinced on protobuf yet (but see below)
+
+- "stop suite" should report to user "waiting for active tasks to finish" (e.g.)
```
matthewrmshin: And:
* report the list of active tasks it is waiting for?
* provide hint to users about other stop options?

docs/meetings/exeter-jun-2019.md
```diff
@@ -0,0 +1,254 @@
...
+   - failed tasks should not be treated differently to succeeded in terms of
+     workflow
+       - instead of keeping them around in the task pool should we have
+          a different failure alerting system?
+        - need to distinguish tasks that are expected to fail
+           - FUTURE: generalise this to allow users to specify "expected
+           completion state" of all tasks (default to expect success)
+   - NOTE two kinds of ghost nodes!
+      1) graph node with no corresponding task proxy currently in the pool
+         (this kind will cease to exist in the spawn-on-demand future).
```
matthewrmshin: Nodes N-generations away from active tasks. Can be N-parents or N-children. Should these be ghost tasks? Or just call them inactive or non-current tasks?

docs/meetings/exeter-jun-2019.md
```diff
@@ -0,0 +1,254 @@
...
+      2) "wrong graphing": intercycle dependence on an undefined task
+    (What did we decide on terminology here???)
```
matthewrmshin: Can we just call these undefined tasks?

docs/meetings/exeter-jun-2019.md
hjoliver: Yes, I'm keen to use it.  Can you "invite" everyone, if you haven't already?

docs/meetings/exeter-jun-2019.md
hjoliver: I think the idea was: Jinja2 variables in `rose-suite.conf` are currently copy-and-pasted into the install suite.rc file.  Instead we should define those variables in a Python module, and import it in the `suite.rc`.  Which is now possible: https://cylc.github.io/doc/built-sphinx/suite-config.html#importing-additional-python-modules

docs/meetings/exeter-jun-2019.md
hjoliver: I think we agreed the proposal though, no?

docs/meetings/exeter-jun-2019.md
hjoliver: I think the proposal was to use `run%d` by default, but to allow users to specify individual run names.  But your idea is good, as an additional option :+1:

hjoliver:
Thanks @matthewrmshin - I will address  your comments by amending the doc soon...

docs/meetings/exeter-jun-2019.md
kinow: Funny, I thought I had used Zenhub in Exeter, but tried to log in with GitHub, and it's asking me to grant the permission as if it was the first time. Scary part is that it requests read/write for all public and private repositories :/

hjoliver: I guess it needs write permission so that it can automatically create Issues from cards etc.

matthewrmshin: I have now invited all of us in the UK and NZ teams.

docs/meetings/exeter-jun-2019.md
matthewrmshin: I suppose we can also enhance the `--set-file=...` option of the `cylc run` command to do this?

docs/meetings/exeter-jun-2019.md
matthewrmshin: I have written the following in the Trello board. Probably lost in the ether now.

Considerations:
* How to easily start multiple runs with multiple versions (or tree-ish objects)?
* How to easily start multiple runs with multiple configuration sets?
* How to easily start multiple runs with multiple ad-hoc changes?
* Where are the settings stored? How to present these to users?
* Version control of configuration on run and reload.
* How to apply changes to source to all current runs?
* Configurable run names.
* Plugin to improve UX for version controlled source.
* Multiple sources. Source direct from a VCS or online resource?
* Managing user expectation on any major changes.
* If a suite is considered a normal software project, running it should be considered an installation/deployment issue like any other software project.

docs/meetings/exeter-jun-2019.md
matthewrmshin: (See also #1032. Reason for shutdown should be recorded.)

dwsutherland:
Not sure the exeter notes are editable, or more historical...

I've already discussed some of the questions/features, but just thought I would responed to points relevant to me:

> - GraphQL has typing too (like Protobuf); not sure  if validation/checking is done
>  - BK wants types in the UI for safety/robustness

As discussed with Bruno, GraphQL does request/query validation (both field and Arg/Var type checking) and type validation on data returned by the resolvers..
This covers all bases, and means you don't have to follow any bread crumbs in finding errors.. The fact that both protobuf and graphql definitions are in `cylc-flow` makes it quite robust (both WRT development and production).


>GraphQL:
>- note graphene (Python client/server) is much less powerful than >Apollo (JS)

We're not using python client, but would agree Graphene is probably not as "powerful" as Apollo GraphQL Server..
There's technology out there to write language agnostic GraphQL schema, so `cylc-flow` could possibly still share it's schema with a non-python implementation, and with use of a Redis like data store perhaps we could migrate the UIS/Tornado to something like Apollo GraphQL Server... However:
- It would need to play nice with Jupyterhub (unless the UIS is split in two).
- I think it could interact with ZeroMQ? (https://github.com/zeromq/zeromq.js/   ?)
- We would loose other functionality imported from `cylc-flow` (scanning/spawning suites, ZeroMQ server/client code  ...etc)

So not sure this is worth entertaining unless we are suffering in performance after optimisation (or there is other interplay from having Apollo client & server)..

>- request for "hold all of my suites": "workflows/*, status=held"

Already capable... Use the GraphiQL interface docs to find out (or just ask!)

>- (what about: hold all tasks submitting to host X?)

Would need to add an arg for host, and possibly a field to `taskProxy` (and that could be cross/multi-workflow also) ..

>Protobuf discussion (as implemented by DS already):
>- decision: Protobuf is fine for the current model where UIS is >*mirroring* all
> data in the "active view" of WFS.  GraphQL backend potentially >better *if* in
> future we can make the UIS merely a transparent conduit for exactly >what
>subset of the data that the UI requests.

I think I may have spoken to this point. We don't want to place strain on the WS by the GUI (something BOM also stressed) and it becomes highly inefficient/impractical for multi-workflow queries... If you still want GraphQL at the UIS (it not just be a proxy to the WS), in order to unify the CLI we import the schema from `cylc-flow` so you can't have both..

>- still need protobuf feed somewhat granularized to level of nodes >and edges,
>  so UIS doesn't store edges if no one is looking at the graph view.
> - Ask DS not to granularize further than this, if not strictly needed.

Not sure there's a massive benefit from doing this... I think we could definitely only sync when a user is connected.. In the spawn-on-demand era we may have to generate edges and nodes on the fly from the workflow pattern (else looking into the future might be hampered), but we still have to store "what happened" somewhere..

>Future:
>- WFS populate in memory DB (redis etc.) ... then UIS can just query >the DB?

Could be beneficial if directly exposed (like redis), then no need for data store sync with WS => UIS (as they would both use this).

docs/meetings/exeter-jun-2019.md
hjoliver: Yes, I think so.

docs/meetings/exeter-jun-2019.md
hjoliver: Good question.  In the old/current meaning of "ghost" (no task proxy in the pool) some of those will be ghosts and some won't.  So now you want to overload the term again to mean non-active tasks visible when you zoom out in the active window ... he heh  :rofl: :rofl:

Hmm, seriously, I think just "active" and "inactive" will do in the future. Ghost implies something a bit mysterious, which "not present in the task pool" is.  But "active or not" should be much less mysterious to users, I hope.

docs/meetings/exeter-jun-2019.md
```diff
@@ -0,0 +1,254 @@
+# Exeter Meeting Notes
+
+June 24-28
+Present HO, DM, MS, OS, SB, BK
```
wxtim: TP

hjoliver: Very sorry @wxtim - you are in fact exceedingly memorable!

docs/meetings/exeter-jun-2019.md
hjoliver: Yes. Comment noted.

docs/meetings/exeter-jun-2019.md
hjoliver: Noted.

docs/meetings/exeter-jun-2019.md
hjoliver: Noted.

docs/meetings/exeter-jun-2019.md
hjoliver: Noted.

docs/meetings/exeter-jun-2019.md
hjoliver: Noted.

docs/meetings/exeter-jun-2019.md
hjoliver: Noted.

docs/meetings/exeter-jun-2019.md
hjoliver: Noted.

hjoliver:
@dwsutherland - pasted  your comments in.

docs/meetings/exeter-jun-2019.md
hjoliver: Noted.

hjoliver:
Feedback incorporated; merging.

PR #40, opened 2019-07-24
Planning Work: Extend Cylc Run (Replace Rose Suite Run)

wxtim:
- [x] Roadmap for small PRs to a dev branch added.
- [x] Added a basic example suite.rc file
- [x] Proposed changes to naming conventions
- [x] Clarify naming conventions
- [x] Clarify tasks to be undertaken
- [x] Move task list near to top of plan doc, or to a separate document
- [x] Create issues for each identified task.
- [x] Expand on discussion of rose-suite.conf deprecation.


docs/proposal-rose-suite-run.md
```diff
@@ -84,11 +84,58 @@ configuration logic. Some points to consider:
...
+            class = slurm
+            before command = sshare -u %(user_id)
+            after command = sacct -j %(jobid)s --format=""
+    [[xcs]]
+        login hosts = xcslr0, xcslr1
+        retrieve job logs = True
+        [[[batch system]]]
+            class = pbs
+        [[[default directives]]]
+            --some-directive
```
matthewrmshin: Cylc configuration does not support this syntax at the moment. We'll need to add an `=` at the end. I wish we can support *list*.

docs/proposal-rose-suite-run.md
```diff
@@ -84,11 +84,58 @@ configuration logic. Some points to consider:
...
+        login hosts = vld392
+
+[runtime]
+    [[Alice]]
+        ...
+
+    [[Bob]]
+        ...
+        [[[platform]]]
+            platform = desktop
```
matthewrmshin: I think we'll merge the old `[[[remote]]]` into the `[[[job]]]` section. So:

```
[[[job]]]
    platform = ...
```

docs/proposal-rose-suite-run.md
```diff
@@ -149,3 +196,38 @@ a single command (e.g. `cylc flow`?) to:
...
+- [ ] Identify which tests should work in a different way
+  - [ ] Tests currently in the Rose Codebase
+  - [ ] Tests currently in the Cylc Codebase
+- [ ] Create a development branch of master for this set of
+  changes.
+- [ ] [PR] Agree modifications to tests which will check
+  that these changes have worked.
+- [ ] Identify aspects changes not covered by tests
+  - [ ] Create tests or explain why tests cannot be created
+- [ ] [PR] Implement the new `suite.rc` & `flow.rc` parsing
```
matthewrmshin: Doing the rename as part of this change is not necessary, but will make some sense.

We should call all of these files `flow.rc` (or `cylc-flow.rc`) and they will share the same schema. The suite `flow.rc` will use the full schema. The site/user `flow.rc` will use a reduced version of the same schema?

docs/proposal-rose-suite-run.md
```diff
@@ -149,3 +196,38 @@ a single command (e.g. `cylc flow`?) to:
...
+  provides the downstream program with desired variables.)
+  - [ ] Bring the `SPEC` dictionaries in `suite.py` &
+    `globalcfg.py` into line, at least for the variables
+    under the `job_platforms` section.
+  - [ ] Deprecate settings for job host
+  - [ ] Assert that any tests referring to parsing of
+    `suite.rc` and `flow.rc` pass.
+- [ ] [PR] Implement actual use of the new `job platforms`
+  - More Detailed points? Are we implementing this in
+    `scheduler.py`?
```
matthewrmshin: Most job logic are under `cylc.flow.task_job_mgr`.

docs/proposal-rose-suite-run.md
wxtim: Sounds very sensible

docs/proposal-rose-suite-run.md
wxtim: Will we force the reduction of that schema? Or would the same schema be available to both if for some weird reason a site wanted to add a default suite graph (say) to their `global-flow.rc`. 

I had something like this in mind:
flow_rc_settings_locs

hjoliver: He's not suggesting that, according to my reading.

hjoliver:
> We should call all of these files `flow.rc` (or `cylc-flow.rc`) and they will share the same schema.

I prefer `cylc-flow.rc` ... to leave no ambiguity about the file's purpose in life.

docs/proposal-rose-suite-run.md
```diff
@@ -24,27 +22,75 @@ The logical step is therefore to move these extra functionalities provided by
...
+  command and command args.[#1030](https://github.com/cylc/cylc-flow/issues/1030)
+  - [ ] Summarize debate on ticket.
+  - [ ] Create a PR against cylc-admin creating a specification for `cylc flow`
+
+
+- [ ] Create a `cylc flow` CLI.
+  - [ ] Include suite validation in `cylc flow` CLI.
+
+
+- [ ] Replace `rose-suite.conf`.
```
wxtim: I feel like the bottom half of this list is increasingly sketchy. I'd be happier of people with bigger domain knowledge might care to colour it in a bit...

docs/rose-suite-run-proposal/future-cli-conventions.md
```diff
@@ -0,0 +1,42 @@
+Placeholder - populate in new PR
```
wxtim: It is not anticipated that this is the finished form of the document: One of the tasks on the work plan is, after all, to summarize the debate in this document in another PR.

oliver-sanders:
Should include something in this document about registering runs under the name of the template suite as discussed in June e.g.:

```
mi-aa001/
    run1
    run2
```

docs/proposal-rose-suite-run.md
```diff
@@ -24,26 +22,73 @@ The logical step is therefore to move these extra functionalities provided by
...
+  - [ ] Create an alternative jinja2 format and tests for this format.
+  - [ ] Create a rose-suite.conf parsing plugin for cylc flow.
```
wxtim: @TomekTrzeciak Suggests that we might prefer to use a python file rather than a jinja2 file as per PR #3169

oliver-sanders: I think we were keen on this approach when last discussed.

hjoliver: Yes, discussed in June, as recorded in my meeting notes (not sure who was present at the time): https://github.com/cylc/cylc-admin/pull/37/files#diff-68ceda0f80a6e03c8997fecc75329aafR50

oliver-sanders: note: we will still need to support the `rose-suite.conf` for the foreseeable future (or provide a converter?).

wxtim: Yes, a converter is to be provided. Will add to list.

sadielbartholomew: Rose configuration is something that I have been assigned as a main piece of work for now onwards, & I believe this step will come under the remit of my work on that, rather than being part of your ``rose suite-run`` work.

How & when we start to migrate encapsulated suite-level configuration from Rose into Cylc, & how it will fit in, is something that definitely needs a whole discussion on its own, I think. I can't see an Issue for it under ``cylc-flow`` or ``rose``, is anyone aware of one that I may have missed?

matthewrmshin: This here is to ensure that `cylc run` (or whatever) @ Cylc 8 is able to read and process an existing Rose suite (with its `rose-suite.conf` and optional configuration). The work here is to:
* Open up a plugin interface for running 3rd party logic on suite start up.
* Write a plugin that uses the Rose library to interpret and act on suites that still have a `rose-suite.conf` file.

It should not impact on the work on `rose config-edit` - except that the plugin may get to use any new API that will be developed for the `rose config-edit` work.

I guess the wording here should be changed to emphasis the **plugin** bit to avoid confusion?

sadielbartholomew: Ah, I see. Thank you @matthewrmshin for clarifying that.

> I guess the wording here should be changed to emphasis the plugin bit to avoid confusion?

Yes, :+1:, a wording change would be helpful, as I clearly interpreted this differently to the intended purpose.

TomekTrzeciak: For completeness, the PR mentioned is cylc/cylc-flow#3169. In its simplest implementation cylc could just check on suite start for some fixed-name file, like say `suitevars.py` that could be overridden with `--set-file` option. I think this would already provide a lot of the required functionality.

One thing that I would change in that PR is the order in which options are processed and handle `--set FOO=BAR` before (and passed it to) `suitevars.py` or `--set-file=...` - this would allow for conditional logic within the python file based on some command line values. Also, it would be useful to have the suite name (as seen by Cylc for suite registration) passed to this python file as well, so one could configure the suite differently depending on its name (we tend to deploy multiple suites derived from a single source base - it would make it possible to have self-configurable suites with minimal options).

wxtim:
Can I please ask @oliver-sanders  and/or @sadielbartholomew  and or @matthewrmshin  to formally review this plan. I think I should get it in, especially as it's a little bit meta (it calls for further planning of the new names for cylc ???? and the replacement .rc files.)

docs/proposal-rose-suite-run.md
```diff
@@ -24,26 +22,81 @@ The logical step is therefore to move these extra functionalities provided by
...
+    to be available in the combined `flow.rc`. [Link to it](example_files/flow.rc)
+  - [ ] Deprecate settings for job host?
+  - [ ] Add sections for `[job platforms]`
+
+
+- [ ] Implement new `cylc-flow.rc` form. [Cylc-flow #3260](https://github.com/cylc/cylc-flow/issues/3260)
+  - [ ] Check old tests for `global.rc` & `suite.rc` to ensure that functionality
+    is not lost.
+  - [ ] Devise tests for the new flow.rc
+  - [ ] Create new config parser, perhaps called `cylc-flow-rc-parser.py`
```
oliver-sanders: Wait what! You do not want to re-write the parsec config parser.

docs/proposal-rose-suite-run.md
```diff
@@ -24,26 +22,81 @@ The logical step is therefore to move these extra functionalities provided by
...
+  - [ ] Summarize debate on ticket.
+  - [ ] Create a PR against cylc-admin creating a specification for `cylc flow`
+
+
+- [ ] Create a `cylc flow` CLI.
+  - [ ] Include suite validation in `cylc flow` CLI.
+
+
+- [ ] Replace `rose-suite.conf`.
+  - [ ] Create an alternative jinja2 and/or python format file and tests for
```
oliver-sanders:
```suggestion
  - [ ] Create an alternative python config file and tests for
```

docs/proposal-rose-suite-run.md
```diff
@@ -24,26 +22,81 @@ The logical step is therefore to move these extra functionalities provided by
...
+    these format.
+  - [ ] Create a cylc plugin to maintain back compatibility by parsing
```
oliver-sanders: FYI: We want to create a formal system for plugin "hooks" in Cylc Flow so this work can piggy-back on that.

https://github.com/cylc/cylc-flow/issues/2959

docs/proposal-rose-suite-run.md
```diff
@@ -136,16 +236,28 @@ sources on installation. Other things to consider:
 The `rose suite-run` command calls `cylc validate --strict` by default.
 Automatic suite validation should become the default behaviour for `cylc run`
 and `cylc reload`. Things to consider:
-* Is the *strict* mode necessary by default?
-* `rose suite-run` does not currently validate Rose items. This should be
-  changed to ensure that all aspects of the suite are validated uniformly.
+* `rose suite-run` does not currently validate Rose apps against their
+  metadata. Rose apps should be validated against their metadata by default,
```
oliver-sanders: To clarify are you suggesting a Rose plugin to Cylc which validates all applications during the suite install process?

Conceptually sounds good to me but we may need to consult users before considering this as it is potentially controversial.

Rose has different levels of validation:

* Built-in (fast)
* Fail-if Warn-if (fastish)
* Macros (slow)

What level would we validate to? Would we validate optional configs?

We might want to upgrade `rose macro` to use asyncio or multiprocessing.

docs/proposal-rose-suite-run.md
```diff
@@ -136,16 +236,28 @@ sources on installation. Other things to consider:
...
+  but contain switches in the manner of bandit or pylint in a config file.
```
oliver-sanders: Not so sure about this, If the config doesn't fit the metadata, either turn off the metadata or fix the config.

docs/proposal-rose-suite-run.md
```diff
@@ -136,16 +236,28 @@ sources on installation. Other things to consider:
...
-* Restart a suite.
-* Restart a suite at a given checkpoint.
-* Warm start a suite.
-* Cold start a suite.
-* Clean start a suite.
+a single command (e.g. `cylc flow`, `cylc go`?) to with
+appropriate command switches should be agreed on in a pull
+request to this repository modifying [this document](rose-suite-run-proposal/future-cli-conventions.md)
+
+## Replace Rose Fileinstall & rose-suite.conf
```
oliver-sanders: This is already outlined above.

docs/proposal-rose-suite-run.md
wxtim: Quite.

docs/proposal-rose-suite-run.md
wxtim: I believed this to be the case, and am glad indeed to have it confirmed

docs/proposal-rose-suite-run.md
wxtim: I believe that is the general plan, but I agree, there is the possibility of some controversy. These are all good points, and I have amended the document to reference them.

docs/proposal-rose-suite-run.md
wxtim: I've re-worded this, and hope it fits a little better.

docs/proposal-rose-suite-run.md
wxtim: Yay! Deletions

docs/proposal-rose-suite-run.md
oliver-sanders: I'm not sure switches are a good idea.

wxtim: Why not?

wxtim:
I'm going to merge this: It's a working document under version control. Thank you all for your help on this - no doubt there will be more to come....

docs/proposal-rose-suite-run.md
oliver-sanders: > Why not?

1. Because I can't really think of legitimate use-cases.
2. Because people might use them!
3. For one off exceptions we should provide a `--no-validate` option or something of that ilk.

*Longer explanation*

Rose allows people to specify metadata and populate configurations. We then validate the configurations against the metadata to see if it fits. If it doesn't either the configuration is broken and needs fixing or the metadata is out of date and needs fixing.

If we give people switches they can hardcode in their configuration to turn specific checks off is, I think, kinda a bad idea. These switches would be like exceptions and are kinda an extension of the metadata itself but hardcoded in the configuration. It would encourage divergence of metadata and configurations. It's kinda OK when its your little hack because you couldn't be bothered to update the metadata when you were fiddling with some setting, but when that gets committed to the repository someone else will inherit this hack. It might even set the precedent this this is how you get `rose macro -V` to pass.

wxtim: Thank you Oliver! I got this from you in person but I like that  you wrote it up for completeness.

PR #41, opened 2019-07-25
Add CLI to workflow service authentication proposal

sadielbartholomew:
I would like to start work on the authentication between the command line & the workflow service, but first we need to agree on the methodology.

The Met Office team have discussed this in a team meeting & come to an agreement on a proposed approach, so I have captured this in a document that I am raising to register with you @hjoliver, as we hope to discuss it in next week's video conference. Could you please have a read through in advance & also add this as an item to the VC agenda?

Only one reviewer necessary, but I have assigned @matthewrmshin too should he want to read through my interpretation of our discussion.

~~(I've also created a 'proposals' directory in this PR, where this new documents sits & where I moved the 'rose suite-run' proposal document too, after realising there are a lot of files currently sitting in the ``docs`` dir & trying to collect some into a sub-dir that is likely to be applicable to future documents.)~~ I'm reverting this aspect now as it is not important & Tim is editing this document in parallel, so I would be stepping on his toes.

Thanks.


docs/proposals/proposal-workflow-authentication.md
```diff
@@ -0,0 +1,69 @@
+# Proposal for Workflow Authentication
+
+## Proposal summary
+
+**For workflow-to-workflow-service authentication, use:**
+
+* **for jobs: one-time per-job tokens**, i.e. a single one-time token for
+  every job;
```
matthewrmshin: for the life time of the the task-job?

docs/proposals/proposal-workflow-authentication.md
```diff
@@ -0,0 +1,69 @@
...
+* **for the user/interactive CLI: timed tokens**.
```
matthewrmshin: For the same logged in user with full access to the file system where the WFS is running as the same user.

Out of scope: Future CLI commands that are not invoked under the above condition may go through the UIS.

docs/proposals/proposal-workflow-authentication.md
```diff
@@ -0,0 +1,69 @@
...
+
+Per-job tokens would require one of the following (or another alternative):
+
+  * another file per job, which is not ideal as it adds to the file count;
+  * a shared combined file containing tokens for all jobs per workflow, also
+    not ideal as it means that jobs would needlessly know about other jobs'
+    tokens;
```
matthewrmshin: A dumped hash table is not so different to the above, because job has access to other user files in the file system any way.

docs/proposals/proposal-workflow-authentication.md
```diff
@@ -0,0 +1,69 @@
...
+solutions:
+
+* "one-time" tokens: single-use tokens i.e. ones that are created then deleted
+  once they are no longer required, with fine (process) granularity for
+  creation/deletion;
+* "timed" tokens: tokens that expire after a set time period.
+
+Due e.g. to the new architecture, job clients must be distinguished & treated
+differently from user (UI & interactive CLI) clients, unlike for the
+[simple passphrase-based approach of Cylc 7](cylc-7-architecture#authentication).
```
matthewrmshin: The *passphrase* in the old system is really a suite-level one-time-token as it is generated randomly, and is valid for the lifetime of the suite. It is not a mnemonic passphrase provided by a human user.

docs/proposals/proposal-workflow-authentication.md
```diff
@@ -0,0 +1,69 @@
...
+
+We also want to make sure the new approach offers increased security, rather
+than just being a "rebrand" accounting for the above, to the Cylc 7 approach.
+
+
+### Difficulties & complications
+
+We need to (sooner or later) consider the following points:
+
+* Jobs can run where there is a non-shared filesystem;
```
matthewrmshin: Is one-time token only relevant to remote jobs in non-shared file system? Jobs running on shared file system with the suite will have access to the local token any way?

docs/proposals/proposal-workflow-authentication.md
```diff
@@ -0,0 +1,69 @@
...
+* We should consider the longer-term future, where jobs are likely to run on
+  cloud platforms;
+* Later work on authorisation may rely on a mapping against tokens, so
+  we should not preclude that;
+* There will be an overlap between the new one-time & timed token approach &
+  the old passphrase approach that needs to be managed.
```
matthewrmshin: Unlikely?

docs/proposals/proposal-workflow-authentication.md
```diff
@@ -0,0 +1,69 @@
...
+
+
+### An alternative to consider
+
+Another idea that was raised was to instead have one token *per host* for
+job clients. Different platforms have different security levels & a
+per-platform approach could benefit from that.
```
matthewrmshin: A timed token per remote job platform.

hjoliver:
@sadielbartholomew - 

> I would like to start work on the authentication between workflows & the workflow service,...

That's good news, thanks!  However, presumably you mean "between Workflow Service and UI Server"?

docs/proposals/proposal-workflow-authentication.md
```diff
@@ -0,0 +1,69 @@
+# Proposal for Workflow Authentication
+
+## Proposal summary
+
+**For workflow-to-workflow-service authentication, use:**
```
hjoliver: As per my other comment, the terminology needs adjusting? ("workflow-to-workflow-service"??)

docs/proposals/proposal-workflow-authentication.md
```diff
@@ -0,0 +1,69 @@
...
+
+
+## Further information
+
+### Context
+
+Cylc 8 discussions have converged (as described briefly
+[here](cylc-8-architecture#command-line-interface) &
+[here](cylc-8-tasks#general-authentication-issues-sujata-hilary--damian-and-martin-)
+on usage of tokens for authentication between any workflow & the workflow
```
hjoliver: Again, I'm a bit confused by the terminology here "any *workflow* & the workflow service"??

docs/proposals/proposal-workflow-authentication.md
hjoliver: It can actually be a mnemonic passphrase provided by the user (if you do that, the suite server won't overwrite it) - but no one ever does that!

kinow:
Hi @sadielbartholomew ! I only spent like 30 seconds quickly reading parts of the text and some comments here. Nothing to add now, except that for authentication I think diagrams (https://github.com/cylc/cylc-doc/issues/37 🎉 ) will be very helpful to quickly understand how things are planned to work.

docs/proposals/proposal-workflow-authentication.md
hjoliver: We also talked about "event-expired tokens" in our Exeter meetings. The token would be expired *by the job finishing*.  But presumably this is what you actually mean by "one time tokens" because every job necessarily uses the CLI (for messages and potentially otherwise) multiple separate times (not "one time")?  If so, maybe we should call them **one job tokens**?

hjoliver:
@sadielbartholomew - nice proposal; all good if amended slightly based on @matthewrmshin and my comments.

matthewrmshin:
Something like this for the simple CLI => WFS case. (Will be a bit more complicated for the WFS => Remote Job => WFS.)

CLI =_ WFS

hjoliver:
Nice diagram :+1:

matthewrmshin:
(This demonstrates a one-time token sequence. (And I have Forgotten to activate the token to demonstrate its lifetime.))

matthewrmshin:
The timed token will start with the WFS => creating token on file system ...

docs/proposals/proposal-workflow-authentication.md
matthewrmshin: Yes. A token that is only valid for the duration of a job.

matthewrmshin: > It can actually be a mnemonic passphrase provided by the user (if you do that, the suite server won't overwrite it) - but no one ever does that!

Don't tell anyone!

docs/proposals/proposal-workflow-authentication.md
sadielbartholomew: Yes, good point, that's an important aspect to clarify. Thanks.

docs/proposals/proposal-workflow-authentication.md
sadielbartholomew: Oh, sorry, this one was actually my interpretation of something you told me! I must have mid-interpreted? Can you recall saying anything along that line & if so help me figure out what you did mean (it might not be important for this document, but for my understanding)?

docs/proposals/proposal-workflow-authentication.md
sadielbartholomew: Yes, it seems "token" is the more appropriate word strictly but I stuck with passphrase since it is the standard terminology in the old approach.

> suite-level one-time-token ...  valid for the lifetime of the suite

Excellent way to summarise the old method, I'll add that in as it helps to make the difference in that & the proposed new approach very clear.

docs/proposals/proposal-workflow-authentication.md
sadielbartholomew: True, I'll flesh out that point a bit.

docs/proposals/proposal-workflow-authentication.md
sadielbartholomew: Sure :grimacing: I suspect I have misunderstood some important aspect here as I am not sure how the UI Server comes into this, I suspect I should not have included the UI as part of the user client scope below & then the topic would just boil down to the CLI for both job & user clients.

I'll look at the comments below & get further clarification from @matthewrmshin today. Sorry for the confusion!

hjoliver: No, I initially thought you meant "WFS to UIS" but that was before I read your document.

I think you actually mean "CLI to WFS" (for CLI commands invoked by jobs or the suite owner, direct to WFS without going via the UIS).

The confusing bit in your description is "**workflow** to workflow service" - what is "workflow" (in bold) here?  I think you mean "job or owner CLI".

sadielbartholomew:
Ooh excellent diagram @matthewrmshin, thank you that captures a lot & helps my understanding. Depicting it visually is a good way to go.

> Will update tomorrow.

Once updated can I pop it in the proposal document here? I'm not sure at what stage you were intending will flesh out the detail of the remote job aspect, but it might help to do, at least as a first draft sensible proposal, in a similar diagram, to addition to this one, before the video conference, then we can discuss it.

matthewrmshin:
Maybe something like this. (Should probably add a destroy token action at WFS near the end.)

wfs= gt;remote-job= gt;wfs

sadielbartholomew:
Feedback addressed, thanks all. I just need to agree on & check the diagrams & get those added in, then this will be ready to go.

sadielbartholomew:
Note to self: there's a few typos & some links are broken due to requiring a '..' specification.

sadielbartholomew:
Okay, this, the finished basic proposal document, should be ready now for final review, looking to merge.

Once I have had some more time to think about it, I will add to the document some detail RE a rough plan to build up to the proposal from the current basic system in place in the codebase (involving JSON Web Tokens) & RE proposals for the tokens, e.g:

* how they will be stored;
* what form they will take.

But I can add that in new sections later. Can we merge this on its own for now, as the information to come will be self-contained in new sections, & I have might be busy with other aspects for a few days?

FYI, I added some initiating interactions to the diagrams by @matthewrmshin & colour-coded the arrows to distinguish some interaction types, which helps (me at least) to comprehend some of the processes occurring in terms of the larger picture. I hope they changes seem okay. If anyone wants to make edits, I have included the source code (to be input into https://sequencediagram.org/) as HTML comments in the markdown doc code here, before the corresponding diagram image include statements.

sadielbartholomew:
**Update:** I've received some useful feedback & further thoughts from @dpmatthews, & from that will make some amendments to the proposal, so consider this on hold again. Thanks.

sadielbartholomew:
I've just put in a commit to make the proposal document clearer & more useful, by first outlining the problem & *then* making a proposal for a solution, instead of having a mix that wasn't too clear in the first instance anyway.

But there are many aspects I still want to tweak, so *please do not review this yet*.

hjoliver:
> But there are many aspects I still want to tweak, so please do not review this yet.

Not a review - but looks very good so far :+1:

hjoliver:
Note older Issue here https://github.com/cylc/cylc-flow/issues/2729 - the job-suite part of it is presumably superseded by this (haven't had a good read of it recently though, and I'm out of time today).

sadielbartholomew:
Good spot @hjoliver, thanks for pointing it out. I'll have a detailed look at it & try to pick out the points that are relevant to this proposal based on the components we have decided on since the comments & diagrams were written up there. Also, I'm not entirely sure yet how much setting up CLI->WFS authentication could or should go towards session management, so I'll think about that.

matthewrmshin:
> Note older Issue here [cylc/cylc-flow#2729](https://github.com/cylc/cylc-flow/issues/2729) - the job-suite part of it is presumably superseded by this (haven't had a good read of it recently though, and I'm out of time today).

Yes, this proposal supersedes the **WFS <=> Job** proposal in cylc/cylc-flow#2729.

hjoliver:
Where are we at with this PR now @sadielbartholomew ?  Have you finished updating it since our VC meeting?

sadielbartholomew:
Sorry for not clarifying @hjoliver, I updated it since the VC but it is not ready to merge yet as I wanted to make sure Dave was happy (having been on leave for the VC of course) with the otherwise-agreed proposal & he has some concerns.

We are going to discuss those as a MO team to agree between us on a tentative suggestion to bring back to you & the wider team for amendments to that plan, else on some way to discuss it to get to a stage of agreement (so everyone co-located can come to some consensus to minimise debate required online). That is scheduled for tomorrow (Friday) morning (UK time), so I will not touch this until then & in the afternoon after that discussion I will comment on here or Riot to convey our thoughts & how I might want to edit the proposal in light of that.

So in short, consider this 'not ready' until you hear otherwise. I want to make sure everyone is agreed on a plan & that the proposal indicates the final working plan when it gets merged (so perhaps 'proposal' isn't the right word anymore!).

hjoliver:
OK, no worries :+1:

sadielbartholomew:
> We are going to discuss those as a MO team to agree between us on a tentative suggestion to bring back to you & the wider team for amendments to that plan, else on some way to discuss it to get to a stage of agreement ...
> That is scheduled for tomorrow (Friday) morning (UK time)

To get back to you as promised, @hjoliver, we had the discussion as outlined, though it was in fact focused mostly on a new security aspect brought to light (& how it relates to this): see #3298.

In short, we need to address WFS communications encryption at some point, & the method we propose to do so could work well for some or even all of the CLI => WFS authentication too, so we think it might be a good idea to hold decisions here on the auth side until we can code & profile a basic implementation for the WFS comms, to help us understand how viable that could be as a solution for the auth.

Would you be happy with that general plan? (I'll flag this on Riot, so feel free to discuss there if it seems more appropriate). @cylc/core, please jump in with your thoughts too!

If there's agreement there, I would then suggest I add a note on this proposal document covering the above & referencing that Issue, & then I suggest we merge this as a tentative plan put on hold, so a proposal, true to its name!

hjoliver:
Yep, sounds good.

sadielbartholomew:
> I would then suggest I add a note on this proposal document covering the above & referencing that Issue (& also making some minor tweaks based on some more feedback from Dave), & then I suggest we merge this as a tentative plan put on hold

Done! As per my previous comment, this is now in a state to merge, where it has been noted in the document as a "plan put on hold", with links to relevant comments.

sadielbartholomew:
Just to bump you in case this is lost in the noise @hjoliver, this should be ready now, so please review at your convenience. Thanks.

PR #44, opened 2019-08-01
Rose suite run proposals cylc flow.rc

wxtim:
Follow on from #40.
- [x] Create a specification document for the new `cylc-flow.rc` file.

PR #45, opened 2019-08-01
Rose suite run proposals changes cylc flow app

wxtim:
Create a specification for the behaviour of the CLI which will replace `rose-suite-run`.

n.b. Follow on from #40. The main document changed in this one is the proposal for the cylc-flow app.

### DIFF
--- a/docs/rose-suite-run-proposal/future-cli-conventions.md
+++ b/docs/rose-suite-run-proposal/future-cli-conventions.md
@@ -1,14 +1,37 @@
-Placeholder - populate in new PR
 # `cylc flow` CLI proposal
-Refers to discussion in [cylc-flow #1030](https://github.com/cylc/cylc-flow/issues/1030).
 
+## Purpose
+Specify how the Cylc8 command to replace `rose-suite-run` will work.
+
+## References
+[cylc-flow #1030](https://github.com/cylc/cylc-flow/issues/1030).
 
 ## Discussion of possible names for new command
 
+Possibilities include:
+
+| Name | For | Against |
+| --- | --- | --- |
+|`cylc run`||Already in use for another command|
+|`cylc suite-run`|similar to `rose suite-run` -familiar|similar to `rose suite-run` -confusing<br>Quite long|
+|`cylc`|Short<br>Hints at the core-ness of this command to Cylc|Already in use|
+|`cylc flow`|Consistent naming<br>Like nothing currently in use  ||
+|`cylc go`|Short|Confusable with `rosie go`|
+|`cylc execute`/`exec`||Long<br>Too generic|
+||||
+
+For the remainder of this document this command will be referred to
+as `cylc flow`.
+
+
 ## Discussion of functionality available
-* Restart a suite from last checkpoint (`--restart`, `--continue`).
-* Restart a suite at a given checkpoint (`--restart-from`, `--restart-at`,
-  `--continue-from`).
+
+The following functionality may be necessary or desirable:
+
+* Restart a suite from last checkpoint (`--restart`, `--continue`
+  `--resume`).
+* Restart a suite at a given checkpoint (`--restart-from`,
+  `--continue-at`,  `--continue-from`).
 * Restart a suite from an arbitary point (warm start, `--warm-start`,
   `--restart-from`, `--continue-from`).
 * Cool start - start a suite from the beginning without cleaning install or


docs/rose-suite-run-proposal/future-cli-conventions.md
```diff
@@ -1,14 +1,37 @@
...
+## References
+[cylc-flow #1030](https://github.com/cylc/cylc-flow/issues/1030).
 
 ## Discussion of possible names for new command
 
+Possibilities include:
+
+| Name | For | Against |
+| --- | --- | --- |
+|`cylc run`||Already in use for another command|
```
oliver-sanders: This will command will replace `cylc run` so this is just keeping the name the same?

docs/rose-suite-run-proposal/future-cli-conventions.md
```diff
@@ -1,14 +1,37 @@
...
+|`cylc suite-run`|similar to `rose suite-run` -familiar|similar to `rose suite-run` -confusing<br>Quite long|
+|`cylc`|Short<br>Hints at the core-ness of this command to Cylc|Already in use|
+|`cylc flow`|Consistent naming<br>Like nothing currently in use  ||
```
oliver-sanders: against: is a noun not a verb?

wxtim: To be against - surely a verb?

docs/rose-suite-run-proposal/future-cli-conventions.md
wxtim: I'm not sure that this is in fact the plan. I think we may keep the old cylc run for legacy reasons, but I'm going to forward this on to @matthewrmshin  becuase I don't have a convincing answer.

docs/rose-suite-run-proposal/future-cli-conventions.md
sadielbartholomew: Forgive me for being terrible, but:
 
scr-define-flow

sadielbartholomew: ...either way, I personally think ``cylc flow`` is not a good idea. It's already the name of the module, & "flow" doesn't imply (at least to me) start-up, which is the general idea (which is probably what Oliver was getting at?).

I think we should choose something that is some synonym of 'start-up' or perhaps of 'install', or something that implies both if there is such a word out there. Something more like ``cylc go``, which is my favourite there, but I still am not that keen on. Suggestions I have at the moment along those lines are ``cylc ...``

* ``launch``;
* ``start``;
* ``activate``;
* ``advance``;
* ``progress``;

Another one on a different train of thought:

* ``work``, c.f. the new (generic) term workflow & the work directory, i.e. let my suite continue with its work), or similarly even ``workout``.

Could we perhaps get some suggestions along those lines, as you prefer, added into the list of possible names?

matthewrmshin: Or `cylc play`. Play-Pause-Stop like a music player. (This is used by the old SMS.)

But I like `cylc go` a lot. Fewer letters to type.

sadielbartholomew: Yes, actually both of those are nice.

And there's the potential for puns involving suite names for the latter, whether that's a good or bad thing I am not sure (``cylc go away``, ... ``go fish``,... ``go to_sleep_for_goodness_sake_hilary_we_know_its_2am_in_NZ``...)

hjoliver:
Being the most important command name, this is as tricky as naming a band or a child, and probably requires a similar amount of alcohol to get it right.

Going into critique mode...

- launch? :-1: - an ugly sounding word (like "graunch" or "raunch")
- start; :-1: - doesn't fit well with the restart/continue concept
- activate; :-1: - suggests "turn it on" without necessarily doing anything else
- advance; :-1: - what if the workflow is held or waiting on triggers?  (then it's not advancing)
- progress; :-1: - ditto
- work; :-1: ...  hmmm, interesting, doesn't do it for me although I can't seem to articulate why
- workout; :-1: very creative!  but suggests something difficult and sweaty!

and:
- go? :+1: - pretty good
- flow? :+1: - brings to mind job activity flowing through the workflow graph

sadielbartholomew:
We don't seem at agree much at all here @hjoliver 😂 Fair enough! Though some comments make me realise the suggestions weren't good ones (eg. for start, advance, progress).

I'll have another think because I would like to add in some suggestions, as I'm personally not sold on any of the existing options.

That was juts my first thoughts, really.

hjoliver:
I'm increasingly liking `cylc flow`.  After all, it is now the name of the back-end system that `cylc run` (or it's successor!) will start up.  And it **is** a verb, as discussed above.  And it it suggests to me, *make the suite flow* ... which works for both start-from-scratch and continue-from-checkpoint.

kinow:
`flow` will reinforce our rename to Cylc Flow too. But `play`/`pause`/`stop` also sounds good. Would `cylc stop` then stop the flow? :smile:

hjoliver:
`cylc staunch` - as in *to staunch the flow of blood* :rofl:   

It's another 'aunch` word, like I complained about above!

hjoliver:
Maybe we should do play/pause/stop, no one will misunderstand that.  and "play" works for start-from-scratch and continue-from-checkpoint. 

Then we can have fast-forward, and rewind too.  And we should change "workflow" to "mix-tape" :rofl: 

New logo:

image

docs/rose-suite-run-proposal/future-cli-conventions.md
oliver-sanders: Would be happy with play, pause, stop. Would want to symlink "run" to "play" though.

oliver-sanders:
I look forward to a future when the suite db can be easily rolled back to previous states (like implicit checkpoints) giving us the `cylc rewind` command...

oliver-sanders:
Not so keen on `flow` btw, the `cylc flow` command uses Cylc Flow to run your `cylc flow`.

docs/rose-suite-run-proposal/future-cli-conventions.md
wxtim: The UM is probably something by Bach.

wxtim:
> New logo:

Where do I put the pencil in the HPC to make cylc rewind work?

sadielbartholomew:
One new idea, that is more along the lines now suggested: **``cylc on``** (& naturally then ``cylc off`` to stop)

* short & sweet, fitting @matthewrmshin's criteria :smiley:;
* implies that a workflow has been enabled, without implying it is advancing, to fit in with cases where it is held or waiting where that is not true;
* similar to ``play`` but without the connotation that something recreational or done for enjoyment (!) is being started or resumed.
* one definition for "on" from a Google Search of "define on" is "indicating continuation of a movement or action" which is what we seem to be looking for;
* when sounded out, i.e. "Cylc-on", it sounds cool

hjoliver:
> when sounded out, i.e. "Cylc-on", it sounds cool.

Reminds me (sort of) of some ancient history!
```bash
[cylc/cylc-flow]$  git log --date=short --pretty=format:"%an - %ar - %s" | \
                                egrep -i 'name change (to|from)'

Hilary James Oliver - 10 years ago - name change to cylc, begin integrated user interface
Hilary James Oliver - 10 years ago - name change from cyclon to cylon
Hilary James Oliver - 10 years ago - SYSTEM NAME CHANGE TO CYCLON
Hilary James Oliver - 10 years ago - NAME CHANGE TO CYCON
Hilary James Oliver - 11 years ago - docs + name change to sequenz
```

matthewrmshin:
And the name of the next Cylc conference will be... **Cylc Con**!?

hjoliver:
Hmmm... I think I quite like `cylc on/off`.  It is fantastically minimal.  

The only thing more minimal would be `cylc 1/0` - but that is definitely way too esoteric!

I guess we could keep `hold/release` with it.

hjoliver:
It was "Cylon" at the time of my 2009 Met Office POC demo, where some highly respected SMS dude shot me down in flames.

Original series, of course:
image

sadielbartholomew:
> some ancient history!

Oh my, over 10 years ago :scream: Your baby has grown into a child, soon to hit the teenage phase (gulp)!

It's interesting to see some alternative names you were considering (& that there were so many name changes towards the start). Sequenz is the name of a keyboard product, according to a Google search. I'm not sure if that was the case at the time, & that inspired a change from that one!

docs/proposal-rose-suite-run.md
```diff
@@ -24,26 +22,73 @@ The logical step is therefore to move these extra functionalities provided by
...
+    to be available in the combined `flow.rc`. [Link to it](example_files/flow.rc)
+  - [ ] Deprecate settings for job host?
+  - [ ] Add sections for `[job platforms]`
+
+
+- [ ] Implement new `cylc-flow.rc` form. [Cylc-flow #3260](https://github.com/cylc/cylc-flow/issues/3260)
+  - [ ] Check old tests for `global.rc` & `suite.rc` to ensure that functionality
+    is not lost.
+  - [ ] Devise tests for the new flow.rc
+  - [ ] Create new config parser, perhaps called `cylc-flow-rc-parser.py`
```
kinow: I think this new file and other related files could be grouped under a module, something like `cylc.flow.config.parsers`, then just have a new class in there?

oliver-sanders: `s/parser/spec/g`

sadielbartholomew:
> One request before merge: please can you add the documents you have created to the Cylc Admin document listing in index.md?

Actually, I've just noticed that #44 & #40 that you recently merged also add in some documents that are not listed. Can you add in those too? It will help us to keep track of, & have quick access to, the various documents you have written for RSR.

wxtim:
> > One request before merge: please can you add the documents you have created to the Cylc Admin document listing in index.md?
> 
> Actually, I've just noticed that #44 & #40 that you recently merged also add in some documents that are not listed. Can you add in those too? It will help us to keep track of, & have quick access to, the various documents you have written for RSR.

Quite right. My Bad!

sadielbartholomew:
Thanks for adding those in via #50.

> I don't suppose that you could give it the once over and merge - I don't think it's controversial...

Sure, I'm reviewing that PR now.

docs/rose-suite-run-proposal/future-cli-conventions.md
hjoliver: I'll resolve this discussion, as the point will be obvious to any of us reading the document.

docs/rose-suite-run-proposal/future-cli-conventions.md
hjoliver: @wxtim - can you add the above suggestions (maybe with args for and against) to the document, then we'll merge this and propose any further changes by subsequent PRs.

wxtim:
> Then we can have fast-forward

AKA Simulation mode?

hjoliver:
We need to decide what to do with this one.  

It should probably be re-cast as clear documentation of what all the start/restart options do exactly, followed by a list of options for rationalizing the CLI to achieve that more sensibly than we do now.

hjoliver:
For the workshop we can just note what the problems are with the current start/restart CLI, and some ideas for doing it better, if the topic comes up.

wxtim:
Well and truly dated at this point.

PR #46, opened 2019-08-14
Summary notes on Met Office user groups' feedback

sadielbartholomew:
This adds a document containing notes summarising all significant points of feedback raised by "power" users in two user group meet-ups last month & on pages set up by each group. (It also renames the ``research`` sub-directory to ``user-research``, to be more specific & clear.)

Sorry this has taken a while for me to put up; I've been writing these up from my written notes in my spare time & spare time has been short of late!

As standard for this repo, please confirm you are happy for this to go in @hjoliver, & @oliver-sanders I have added you as a reviewer in case you would like to check this for accuracy, e.g. query anything written, or say if you feel anything important has been missed out etc. (but if you don't feel that is necessary, or don't have the time, feel free to unassign or approve).


docs/user-research/MO-user-groups-feedback.md
```diff
@@ -0,0 +1,276 @@
...
+
+
+#### From feedback pages
+
+Notable points (relating to the Cylc UI) from feedback pages:
+
+* [From the agenda notes:] "TSUG noted that this [the Rose replacement] will
+  be of most benefit to users represented here. The *Cylc upgrades are nice
+  to have* but the Rose upgrades will be vital to allow us to get the most
+  out of our trials."
```
hjoliver: Huh.  Well I would emphasise to these users that the "upgrade" aspect of Cylc 8 will be nice but is actually something of a side-effect. The main thing is: *we **have to do this** because Python 2 and PyGTK are **dead**!*

docs/user-research/MO-user-groups-feedback.md
```diff
@@ -0,0 +1,276 @@
...
+NB: those pages contain quite a lot of useful feedback, but not much
+(just the above) relating to the Cylc UI, mainly to Rose config-edit, plus one
+non-UI Cylc matter (which is of relevance to future Python API work):
+
+* Under the section 'User requirements for new suite control system', from MB:
+  "Better support for determining more complex aspects of the graph of a
+  suite. Some aspects of the suite graph of an ideal version of an evaluation
+  suite would be sufficiently complex that using jinja2 is unwieldy, limiting
+  and difficult to maintain. It would be useful to have the ability to define
+  the suite programmatically either as an initial app, or by some other means.
```
hjoliver: Bring on Cylc 9!

docs/user-research/MO-user-groups-feedback.md
```diff
@@ -0,0 +1,276 @@
...
+
+* RG: Will we be able to customise the datetime formats? I find them very
+  unwieldy in their full form. [We explained that it would be difficult &
+  impractical to change the datetimes used internally.]
+* RG: (clarifying previous comment) I don't mean customising them as used &
+  processed for the suite internals, just for the UI display, i.e. what we
+  will see on the front-end. [We stated that we have considered datetime
+  customisation on the UI, & hope to enable some means to make the datetime
+  formats clearer there, for example by including the ':' & '-' delimiters even
+  if not used in the configuration, to break apart the datetime components.]
```
hjoliver: Presumably you mentioned (e.g.) `cycle point format = %Y-%m-%dT%H:%M`?

hjoliver:
@oliver-sanders - you can merge this if you agree with @sadielbartholomew 's summary.

docs/user-research/MO-user-groups-feedback.md
sadielbartholomew: Very true, though sadly I don't think (from my recollection) that we managed to convey that, mainly because we only had a short time slot to discuss feedback & that the mention of Rose gave us a cue to emphasise, as we thought important, that we did want to hear feedback on Rose (just not in that particular session, which was devoted to the Cylc UI). And then discussion flowed from that such that the topic changed before we could describe the requirement to "upgrade".

Hopefully it was somewhat evident from context, though, that there were push factor(s) & we weren't just changing the UI etc. because it would be nice to do.

docs/user-research/MO-user-groups-feedback.md
sadielbartholomew: Absolutely. It was nice to read that & think "we've got you covered there" (in our intentions & motivations for them, at least!)

docs/user-research/MO-user-groups-feedback.md
sadielbartholomew: I can't say for certain, since it's not in my notes & my memory is fading & the meetings were a month ago now, but I suspect it was probably mentioned. Regardless, I am confident the person in question would have been aware of that configuration item, &/or as a Rose user likely know about ``rose date``.

oliver-sanders: > Presumably you mentioned (e.g.) cycle point format = %Y-%m-%dT%H:%M?

Yes.

hjoliver: So do we have this point covered (users can set the date-time format as they like) or do they want us to display formatted date times in the UI regardless of the internal format used?  That may not be a good idea as the date times displayed in the UI won't match those passed to task jobs and used in filenames etc.

sadielbartholomew:
Can I bump you on this @oliver-sanders, so we can get it merged soon. Thanks.

PR #50, opened 2019-08-29
Rose-suite-run migration docs added to index

wxtim:
Mistakenly committed these docs without index refs to them.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -21,6 +21,9 @@
 
 ## Orthogonal Work in Progress
 - [rose suite-run migration proposal](proposal-rose-suite-run.md)
+  More detail on rose-suite-run migration:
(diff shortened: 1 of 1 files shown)


wxtim:
> This looks fine, thanks.
> 
> One other thing, sorry: I've just noticed there is the original proposal document under `docs` but not further still under the `rose-suite-run-proposal` sub-dir. Ideally we can keep everything relating to RSR contained together, so what do you think about moving `proposal-rose-suite-run.md` under that sub-dir too?

My idea was that the overview should remain in the top level folder with docs referred to by it in the sub folder. How strongly do you feel about this?

sadielbartholomew:
> My idea was that the overview should remain in the top level folder with docs referred to by it in the sub folder.

Fair enough, if it is an overview after all (I wasn't sure of it's relation in you the writer's eyes to the other RSR documents) then its current location is fine. 👍

PR #55, opened 2019-09-16
Add threat modeling "stories"

hjoliver:
(I'm just putting this us as a PR to ensure awareness of its existence beyond me; I originally posted it to Riot chat back in late March I think, but there's a danger we'll forget to respond ... hence this; we can merge after a few approvals have accumulated).

(Note we could address most - or at least many - of these points already, but I'm in no hurry to do so as we have so much "real work" to do yet).

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -73,4 +73,4 @@ Guides for developers, and therefore slightly separate to main documentation
 - [Getting Started](howto/environment.md)
 
 ## Misc
-- [thread modeling outcomes](threats.md)
+- [threat modelling outcomes](threats.md)
--- a/docs/index.md
+++ b/docs/index.md
@@ -71,3 +71,6 @@ Agendas and post-meeting notes.
 ## Cylc 8 Guides
 Guides for developers, and therefore slightly separate to main documentation
 - [Getting Started](howto/environment.md)
+
+## Misc
+- [thread modeling outcomes](threats.md)
--- a/docs/threats.md
+++ b/docs/threats.md
@@ -22,7 +22,7 @@ accepted technology for use at the BOM.
 - As an attacker I want to harvest usernames and passwords
 - As an attacker I want to be able to get the network map relevant to the Cylc solution
 - As an attacker I want to hijack current tokens to take over sessions
-- As an attacker I want to harvest suite state via the HTTP proxy on Cylc hub
+- As an attacker I want to harvest suite state via the HTTP proxy on Cylc Hub
 - As an attacker I want to execute my own suite/arbitrary code on HPC
 - As an attacker I want to inject arbitrary code into al already running suite
 - As an attacker I want to 'spawn' a non UI server (arbitrary code, not cylc)
--- a/docs/threats.md
+++ b/docs/threats.md
@@ -24,7 +24,7 @@ accepted technology for use at the BOM.
 - As an attacker I want to hijack current tokens to take over sessions
 - As an attacker I want to harvest suite state via the HTTP proxy on Cylc Hub
 - As an attacker I want to execute my own suite/arbitrary code on HPC
-- As an attacker I want to inject arbitrary code into al already running suite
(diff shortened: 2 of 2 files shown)


docs/index.md
```diff
@@ -71,3 +71,6 @@ Agendas and post-meeting notes.
 ## Cylc 8 Guides
 Guides for developers, and therefore slightly separate to main documentation
 - [Getting Started](howto/environment.md)
+
+## Misc
+- [thread modeling outcomes](threats.md)
```
kinow: s/thread/threat

hjoliver: that was quick :grin: thanks will fix...

docs/threats.md
```diff
@@ -0,0 +1,38 @@
...
+> Security User Stories
+>
+- As an attacker I want to stop user A suite A from running
+- As an attacker I want to be able to fool authentication on the Cylc HUB
+- As an attacker I want to harvest usernames and passwords
+- As an attacker I want to be able to get the network map relevant to the Cylc solution
+- As an attacker I want to hijack current tokens to take over sessions
+- As an attacker I want to harvest suite state via the HTTP proxy on Cylc hub
+- As an attacker I want to execute my own suite/arbitrary code on HPC
+- As an attacker I want to inject arbitrary code into al already running suite
```
matthewrmshin:
```suggestion
- As an attacker I want to inject arbitrary code into an already running suite
```

docs/threats.md
```diff
@@ -0,0 +1,38 @@
...
+how. I expect this set of stories will probably do the same!
+ 
+> The other key outcome of the meeting was that we confirmed that JupyterHub is
+on our Enterprise Architecture team's technology roadmap, which means it is an
+accepted technology for use at the BOM.
+
+> Security User Stories
+>
+- As an attacker I want to stop user A suite A from running
+- As an attacker I want to be able to fool authentication on the Cylc HUB
```
matthewrmshin:
```suggestion
- As an attacker I want to be able to fool authentication on the Cylc Hub
```

Is Cylc Hub the only location to do this?

docs/threats.md
```diff
@@ -0,0 +1,38 @@
...
+- As an attacker I want to harvest usernames and passwords
+- As an attacker I want to be able to get the network map relevant to the Cylc solution
+- As an attacker I want to hijack current tokens to take over sessions
+- As an attacker I want to harvest suite state via the HTTP proxy on Cylc hub
```
matthewrmshin:
```suggestion
- As an attacker I want to harvest suite state via the HTTP proxy on Cylc Hub
```

Is Cylc Hub the only location to do this?

docs/threats.md
```diff
@@ -0,0 +1,38 @@
...
+- As an attacker I want to execute my own suite/arbitrary code on HPC
```
matthewrmshin: And/or other site infrastructure?

docs/threats.md
```diff
@@ -0,0 +1,38 @@
...
+- As an attacker I want to inject arbitrary code into al already running suite
+- As an attacker I want to 'spawn' a non UI server (arbitrary code, not cylc)
+- As an attacker I want to steal SSO from the Cylc Hub
```
matthewrmshin: How is this different from point 2?

hjoliver:
@matthewrmshin - I've committed your simple typo corrections, but this is meant to record the original questions from BOM so I don't think we should "correct" those as such.  So I think your questions above are really the very preliminary response from our side.  Maybe should be added at the end of the document?

matthewrmshin:
No problem!

docs/threats.md
hjoliver: (this should be in our response, rather than a change to the original questions)

docs/threats.md
hjoliver: (ditto - response)

kinow:
Fixing conflicts manually on my local clone and merging it locally too :+1:

PR #65, opened 2019-11-20
Update the rose-suite-run proposal and add a separate platforms proposal

dpmatthews:
Re-write of the rose suite-run proposal. The initial focus of the work will be implementing support for platforms which is now described in detail in a separate document.

### DIFF
--- a/docs/proposal-platforms.md
+++ b/docs/proposal-platforms.md
@@ -1,0 +1,298 @@
+# Proposal - initial support for platforms
+
+This will close [cylc-flow #3421](https://github.com/cylc/cylc-flow/issues/3421)
+and partially address
+[cylc-flow #2199](https://github.com/cylc/cylc-flow/issues/2199).
+
+## Key functionality
+
+* Introduce configuration and logic to recognise compute platforms in place of
+  task hosts. This will resolve a number of issues.
+  * In particular, this allows a compute cluster which can be accessed via
+    multiple login nodes (hosts) to be treated as logical unit.
+  * We will be able to configure multiple platforms which share the same hosts
+    (including `localhost`).
+
+* If a platform has a list of hosts then the host used for each operation on that
+  platform (job submission, polling, log retrieval, etc) will be randomised. This
+  should help balance the load. If a host becomes unavailable, try another random
+  host from the list, until exhausted. This will handle hosts becoming
+  unavailable after a job is submitted which is a major limitation of the current
+  logic.
+
+* It will be possible to specify the platform for a task as a platform alias
+  which defines a list of platforms. For example, you may have 2 different HPC
+  systems (configured as separate "platforms") and you are happy for a task to
+  run on either. The list will be randomised and the platforms tried in turn
+  until successful job submission.
+  * Note that the platform will be selected each time a job is run. If you
+    specify a platform alias which defines a list of platforms, tasks must not
+    assume that retries of that task will use the same platform.
+
+* It will be possible to specify the platform as a command in the same way as we
+  can currently specify hosts as a command. The command must return a valid
+  platform (or platform alias). We hope that this functionality will seldom be
+  needed in the future (the main use at present is to select from a list of login
+  nodes). However, there are still cases where this may be needed. For example
+  running a command which:
+  * Returns the platform which has the shortest queue time.
+  * Chooses the platform according to the day of the week (perhaps you have a
+    platform which can only be used at weekends?).
+
+* Note: it will **not** be possible to specify the platform as an environment
+  variable in the same way as we can currently specify remote hosts as an
+  environment variable (e.g. `platform = $ROSE_ORIG_HOST`).
+  * Unclear why we need this given the Jinja2 variables can be used?
+  * Using `$ROSE_ORIG_HOST` relies on Rose specifying extra options to set this
+    in `[cylc][environment]` which we should avoid if possible.
+
+## Example platform configurations
+
+```ini
+[platforms]
+    [[desktop\d\d|laptop\d\d]]
+        # hosts = platform name (default)
+        # Note: "desktop01" and "desktop02" are both valid and distinct platforms
+    [[sugar]]
+        hosts = localhost
+        batch system = slurm
+    [[hpc]]
+        hosts = hpcl1, hpcl2
+        retrieve job logs = True
+        batch system = pbs
+    [[hpcl1-bg]]
+        hosts = hpcl1
+        retrieve job logs = True
+        batch system = background
+    [[hpcl2-bg]]
+        hosts = hpcl2
+        retrieve job logs = True
+        batch system = background
+[platform aliases]
+    [[hpc-bg]]
+        platforms = hpcl1-bg, hpcl2-bg
+```
+
+Note:
+
+* The default for `hosts` is the platform name and platform section headings can
+  be regular expressions to match multiple platform names. This allows us to
+  define many platforms in a single section and handles the case where you have
+  a large number of separate servers which are identical in terms of their
+  platform properties other than their host name.
+
+* The current logic for finding matching host settings is complicated and not
+  fully
+  [documented](https://cylc.github.io/doc/built-sphinx/appendices/site-user-config-ref.html#hosts-host)
+  (sections defining individual host names take precedence over pattern matches
+  regardless of where they come in the order). We propose a different logic for
+  finding matching platform settings:
+  * The pattern will be matched against the full name using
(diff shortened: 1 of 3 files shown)


docs/proposal-platforms.md
```diff
@@ -0,0 +1,298 @@
...
+  can currently specify hosts as a command. The command must return a valid
+  platform (or platform alias). We hope that this functionality will seldom be
+  needed in the future (the main use at present is to select from a list of login
+  nodes). However, there are still cases where this may be needed. For example
+  running a command which:
+  * Returns the platform which has the shortest queue time.
+  * Chooses the platform according to the day of the week (perhaps you have a
+    platform which can only be used at weekends?).
+
+* Note: it will **not** be possible to specify the platform as an environment
```
wxtim: This is a fairly large chunk of logic that I'll be pleased not to transfer from rose suite-run.

docs/proposal-platforms.md
```diff
@@ -0,0 +1,298 @@
...
+  * Note that the first reference to a section in the global or user config
+    determines its position in the list. You can repeat a section later in the
+    config and alter some settings but this doesn't change its position.
+
+* The names of all new and existing settings are subject to change until we
+  release cylc 8.0. Initially names will be chosen which match current usage
+  wherever possible. We intend to review all the config settings once we have
+  the key functionality in place.
+
+* The default platform is `localhost`. At cylc 7, any settings for `localhost`
```
wxtim: :+1: Now I have a better understanding of this (existing) behaviour I don't like it. It means that there are 3 possible places a default could have originated from. :disappointed:

docs/proposal-platforms.md
```diff
@@ -0,0 +1,298 @@
...
+
+* You can't mix old with new. If `platform` is defined for **any** task then any
+  settings in the `[remote]` or `[job]` sections of **any** task result in a
+  validation failure.
+
+* Any settings in the `[remote]` or `[job]` sections which have equivalent
+  settings in the new runtime section are converted using the normal deprecate
+  method. e.g. `[job] execution time limit`.
+
+* The remaining `[remote]` & `[job]` settings are converted by searched for a
```
wxtim:
```suggestion
* The remaining `[remote]` & `[job]` settings are converted by searching for a
```

docs/proposal-platforms.md
```diff
@@ -0,0 +1,298 @@
...
+  matching platform. If there is none the conversion fails resulting in an
```
wxtim: possibly 
> By searching for a platform which matches **all** the items set.

docs/proposal-platforms.md
```diff
@@ -0,0 +1,298 @@
...
+            batch system = pbs
+        # => platform = hpc (set at job submission time)
+
+    [[delta]]
+        [[[remote]]]
+            host = hpcl1
+        [[[job]]]
+            batch system = background
+        # => platform = hpcl1-bg (set at load time)
+```
```
wxtim: I'd like an example of settings that will fail because the items in `[[[remote]]]` and `[[[job]]]` ≠ any platform. e.g.
```config
# Will fail loudly
[[epsilon]]
    [[[remote]]]
        host = desktop99
    [[[job]]]
         batch system = pbs
```

docs/proposal-platforms.md
dpmatthews: Unfortunately we'll still need to support it for deprecation purposes

docs/proposal-platforms.md
dpmatthews: Fixed

docs/proposal-platforms.md
dpmatthews: Fixed

docs/proposal-platforms.md
dpmatthews: Further examples added

hjoliver:
Two approvals, but I presume @oliver-sanders needs to look over this too?

docs/proposal-platforms.md
```diff
@@ -0,0 +1,312 @@
...
+  environment variable (e.g. `platform = $ROSE_ORIG_HOST`).
+  * Unclear why we need this given the Jinja2 variables can be used?
+  * Using `$ROSE_ORIG_HOST` relies on Rose specifying extra options to set this
+    in `[cylc][environment]` which we should avoid if possible.
+
+## Example platform configurations
+
+```ini
+[platforms]
+    [[desktop\d\d|laptop\d\d]]
```
oliver-sanders: I don't like the use of regexes for section headings due to the difficulties of overriding, however, this functionality will probably only really be used at a site level.

As a bare minimum this would be more flexible:

```ini
[[desktop\d\d, laptop\d\d]]
```

dpmatthews: I think sticking with a single regex is clearer. Hopefully it won't need to be heavily used.

docs/proposal-platforms.md
hjoliver: Not sure I understand your point here. It seems natural to me that some items should have 1) a built-in default - such as batch system defaulting to background - that can overridden by 2) site config, which can be overridden by 3) user global config.  Therefore, for this kind of case the "default" value used by a particular suite could have originated from any of 3 places.  Is this what you mean?  But if you don't like that, what's the alternative? - require every suite to explicitly set every item even when defaults - possibly platform or site specific - are sensible?

docs/proposal-platforms.md
hjoliver: I tend to agree with Oliver here, in that we use comma-separated heading lists elsewhere (e.g. under [runtime] in suite.rc)

docs/proposal-platforms.md
hjoliver: However, if you're only talking about the way we currently use `localhost` settings as a default for (other) job hosts - which I think is the only way this impacts this proposal? - then I think it's OK to force that to be more explicit.

hjoliver:
Merging as-is (but we might want to tweak a few details such as the regex form later...)

hjoliver:
Oh, @dpmatthews already merged it. Duh!  :grimacing:

docs/proposal-platforms.md
TomekTrzeciak: Given that the configs can be templated with Jinja (even the global one), is there a benefit of having regex platform names? Seems like this brings additional rules and logic (which the user must be aware of in some contexts) compared to a simple pick a name from a list or group names under an alias.

In terms of extensions, instead of regex substitution in host names, a more general solution could be to perform runtime (or rather submission-time) templating of settings (https://github.com/cylc/cylc-flow/issues/2415#issuecomment-428552084). This would allow to not only alter the host name in arbitrary ways, but also other settings like job directives, which many users would welcome. I'm not huge fan of Jinja templating, but in this context I think it would work quite well (templating values of settings has much more limited scope than 'freestyle' templating an entire config file).

hjoliver: (@wxtim can you note down @TomekTrzeciak 's suggestion here for consideration, particular the extension to job directives ...)

dpmatthews: The directives bit is already noted in cylc/cylc-flow#2415.

Regarding the regex, it's a trade off. Using a regex allows us to define a large number of platforms in a single section. We could define each platform separately using Jinja but we would end up with a potentially very large config which I don't like the idea of (for instance, I don't want `cylc get-global config` to return a massive long config).

hjoliver: Fair enough, let's stick with regex then, for this.

hjoliver: But, I still think it should be a comma-separated list - to keep each regex simpler, and for consistency with our support for comma-separated lists in other similar contexts.

TomekTrzeciak: Another thought on this. With templating of settings one could cut down on the number of definitions with something like this:

```
#!jinja2
[platforms]
    [[boxNo]]
        host = {% raw %}{{ platform + "%02d" % range(1, 51)|random }}{% endraw %}
```

This would choose at random between `boxNo01` to `boxNo50` every time the host template is evaluated. It goes without saying that with a site installed jinja2 filters arbitrary logic can be performed this way.

hjoliver: Good point (Sorry for serially agreeing with everyone - just no time to consider this in detail right now!).  @dpmatthews  @wxtim @oliver-sanders - we do need to consider templating more widely, and if we have it for other reasons then maybe no need for platform name regexes as well.  Disclaimer: I have not sat down to think through the pros and cons yet.

PR #66, opened 2019-11-27
update to os tickets

wxtim:
My updated version of Oliver's thoughts on the work on Platforms.


oliver-sanders:
Perhaps better to write these up as issues which we can assign / raise PRs against than a static document.

dpmatthews:
> Perhaps better to write these up as issues which we can assign / raise PRs against than a static document.

I think either way is fine. We already have cylc/cylc-flow#3421 to refer to.

hjoliver:
Could @wxtim or @oliver-sanders explain the "work on master" aspects of this?  We don't ever work directly on master (as opposed to feature branches that get merged to master) so I wonder if I'm misunderstanding this.

wxtim:
> Could @wxtim or @oliver-sanders explain the "work on master" aspects of this? We don't ever work directly on master (as opposed to feature branches that get merged to master) so I wonder if I'm misunderstanding this.

Clarification - "working on master" in this context refers to creating small(ish) feature branches and getting these accepted onto master straight off the bat. We're using feature branches to describe a branch on which we work for some time creating subsidiary PR's before merging to master.

PR #68, opened 2019-12-04
Platform aliases don't need extra level

wxtim:
Platform aliases were originally a new subsection each containing 1 item. Why not just have those items in the section above?


hjoliver:
(Retroactively approving!)

PR #70, opened 2019-12-16
Proposal: implement spawn on demand

hjoliver:
From proposal intro:

*We have talked about replacing spawn-next-instance-on-submit with spawn-dependent-tasks-on-demand for some time (years!) but have tended to tie it to Cylc 9 plans such as an abstracted scheduler kernel.*

*However spawn-on-demand is technically orthogonal to these other plans, surprisingly easy to implement in the current system, and it won't make Cylc 9 changes any harder in the future (it will make them easier if anything). The advantages are so great that I propose we seriously consider doing it for Cylc 8*

*...*


hjoliver:
No one objected to this, but there was general approval offline, and there's now a working POC implementation as a cylc-flow PR ... so merging this as reading material for the Feb workshop.

PR #73, opened 2019-12-23
proposal: subscription model

oliver-sanders:
Proposal for how subscriptions could be handled at the UIS.

This is effectively a written form of the following diagram:

graphql

But with subscription examples which are loosely based on reality.

Please comment, feel free to raise larger changes as pull requests against this branch.

### DIFF
--- a/docs/proposal-subscriptions.md
+++ b/docs/proposal-subscriptions.md
@@ -1,0 +1,281 @@
+# Subscription Model Proposal
+
+## The Introduction
+
+Two of the central pillars of the new UI are incremental updates and fine-grained subscriptions.
+
+* Incremental updates involves sending only the data which has changed to the UI, this:
+    * Reduces network traffic.
+    * Reduces the memory footprint of the UI.
+    * Allows us to move away from using a global update loop which makes the UI more responsive.
+* Fine grained subscriptions allow individual Vues (Tree, Graph, Dot, ...) to subscribe to the set o data which they actually need at a particular moment in time, this;
+    * Reduces network traffic.
+    * Reduces the memory footprint of the UI.
+    * Reduces the number of items the UI has to process making updates faster.
+    * Enables the UIServer to dynamically adjust its subscriptions to the Scheduler so as to keep load on the workflow to a minimum.
+
+By reducing the about of data we send to the UI we reduce the load on the browser enabling us to do more with the UI and to make it more responsive.
+
+## The Proposal
+
+1. Individual Vues register and update their subscriptions with a service in the UI *(done)*.
+2. This service "merges" subscriptions together then registers the result with the UIS *(done)*.
+3. The UIS parses and stores these subscriptions.
+4. When deltas arrive from the Scheduler, the UIS loops through active subscriptions to see if there is an overlap.
+   * The "overlap" is the intersection of the delta and subscription fields.
+   * This is event driven as we only need to do it as and when deltas arrive.
+5. If there is an overlap send these fields to the UI which registered the subscription as a delta.
+
+## The Detail
+
+### The Three Data Stores
+
+We have three data stores, at the Scheduler, the UIS and the UI:
+
+* Schedulers each have their own data stores.
+* The UIS has one data store containing the relevant subset of multiple Workflow Servers' data stores.
+* The UI has one data store containing the requested subset of the UIS data store.
+
+All three represent exactly the same data model. The latter two must be kept in-sync with the former.
+
+* The UIS holds a subset of the data from the Workflow Servers.
+* The UI holds a subset of the data from the UIS.
+
+### The Ideal Subscription Interface
+
+A subscription look something like this:
+
+```
+subscription {
+  a {
+    b
+    c { 
+      d
+      e
+    }
+    f
+  }
+}
+```
+
+Deltas look something like this:
+
+```
+{
+  b: 1,
+}
+
+```
+
+```
+{
+  b: 1,
+  c {
+    d: 2
+  }
+}
+```
+
+A subscription is a `set`, a delta is a `dict` which is a subset of the subscription.
+
+### The UI
+
+#### Similar But Different Subscriptions (1)
+
+The UI consists of different Vues which have their own subscriptions.
+
+For example a simple Tree Vue might have the following subscription:
+
+```
+subscription {
+  workflows(ids: [<id>]) {
+    taskProxies {
+      name
+      status
+      firstParent {
+        id
+      }
+    }
+    familyProxies {
+      id
+      name
+      firstParent {
+      }
+    }
+  }
+}
+```
+
+Whereas an un-grouped Graph Vue might have this subscription:
+
+```
+subscription {
+  workflows(ids: [<id>]) {
+    taskProxies {
+      name
+      id
+      status
+    }
+    nodesEdges {
+      nodes {
+        id
+      }
+      edges {
+        source
+        target
+      }
+    }
+  }
+}
+```
+
+These subscriptions are different but they have a large overlap. If we were to register both subscriptions with the UIS we would end up with two similar but different data stores in the UI. This would be in-efficient and opens up discontinuity issues.
+
+Users may want to open many similar but different views on the same workflow so we need to be able to handle this in an elegant manner.
+
+#### Merging Subscriptions (2)
+
+One solution to this problem is to "merge" subscriptions together.
+
+Subscriptions are effectively "sets" of data fields, two subscriptions can be "merged" by taking their union. For example the union of the previous two subscriptions is this:
+
+```
+subscription {
+  workflows(ids: [<id>]) {
+    taskProxies {
+      name
+      id
+      status
+      firstParent {
+        id
+      }
+    }
+    familyProxies {
+      id
+      name
+      firstParent {
+      }
+    }
+    nodesEdges {
+      nodes {
+        id
+      }
+      edges {
+        source
+        target
+      }
+    }
+  }
+}
+```
+
+As an added bonus we can query the UI to find out exactly what it is subscribing to and exactly which view has asked for what data making optimisation/debugging much nicer.
+
+### The Workflow Server
+
+When events occur the Scheduler sends "deltas" containing only the changed information.
+
+For example if a job has changed state from submitted to running we might expect a delta of the form:
+
+```
+taskProxies {
+  id: <id>,
+  status: "running"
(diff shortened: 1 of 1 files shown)


hjoliver:
Thanks for writing this up @oliver-sanders .  As well as informing the still-to-do UIS-UI incremental update implementation, this can also eventually become our definitive developer docs for how the workflow data passes through the whole system :+1:

docs/proposal-subscriptions.md
```diff
@@ -0,0 +1,280 @@
...
+</table>
+
+## Caveats
+
+* The comparison of delta and subscription fields will need to be fast.
+  * Else the UIS will become CPU heavy and sluggish.
+  * There may be potential for caching results.
+* GraphQL interface does not provide a mechanism for deltas out of the box.
+  * We may have to extend / implement this ourselves (can we add a `deltaSubscribe` type to GraphQL?).
+  * If we did change the behaviour of the `subscribe` type 3rd party tools (incl graphiql) would loose the ability to work subscriptions (probably not a problem but worth noting).
```
kinow: I think also `Graphene` is limited to provide always the complete schema requested. So if UI is asking

```
{
a {
  b,
  c
}
```

And only `c` changed, I think Graphene still returns everything, `a`, `b`, and `c`. We may need to either find an option somewhere in Graphene for retrieving only a subset, or modify it (I'm saying Graphene here, but maybe it is done by one of the other graphql dependencies?).

oliver-sanders: Yes, this is the part we would need to solve ourselves.

hjoliver: Via Riot chat, @dwsutherland has reputedly figured out how to do this (and without modifying graphene itself).

docs/proposal-subscriptions.md
```diff
@@ -0,0 +1,280 @@
...
+
+## The Introduction
+
+Two of the central pillars of the new UI are incremental updates and fine-grained subscriptions.
+
+* Incremental updates involves sending only the data which has changed to the UI, this:
+    * Reduces network traffic.
+    * Reduces the memory footprint of the UI.
+    * Allows us to move away from using a global update loop which makes the UI more responsive.
+* Fine grained subscriptions allow individual Vues (Tree, Graph, Dot, ...) to subscribe to the set o data which they actually need at a particular moment in time, this;
```
hjoliver:
```suggestion
* Fine grained subscriptions allow individual views (Tree, Graph, Dot, ...) to subscribe to the set o data which they actually need at a particular moment in time, this;
```

docs/proposal-subscriptions.md
```diff
@@ -0,0 +1,280 @@
...
+    * Reduces network traffic.
+    * Reduces the memory footprint of the UI.
+    * Reduces the number of items the UI has to process making updates faster.
+    * Enables the UIServer to dynamically adjust its subscriptions to the WS so as to keep load on the workflow to a minimum.
+
+By reducing the about of data we send to the UI we reduce the load on the browser enabling us to do more with the UI and to make it more responsive.
+
+## The Proposal
+
+1. Individual Vues register and update their subscriptions with a service in the UI *(done)*.
```
hjoliver:
```suggestion
1. Individual views register and update their subscriptions with a service in the UI *(done)*.
```

docs/proposal-subscriptions.md
```diff
@@ -0,0 +1,280 @@
...
+}
+```
+
+A subscription is a `set`, a delta is a `dict` which is a subset of the subscription.
+
+### The UI
+
+#### Similar But Different Subscriptions (1)
+
+The UI consists of different Vues which have their own subscriptions.
```
hjoliver:
```suggestion
The UI consists of different views which have their own subscriptions.
```

docs/proposal-subscriptions.md
```diff
@@ -0,0 +1,280 @@
...
+
+For example a simple Tree Vue might have the following subscription:
```
hjoliver:
```suggestion
For example a simple Tree View might have the following subscription:
```

docs/proposal-subscriptions.md
```diff
@@ -0,0 +1,280 @@
...
+      id
+      name
+      firstParent {
+      }
+    }
+  }
+}
+```
+
+Whereas an un-grouped Graph Vue might have this subscription:
```
hjoliver:
```suggestion
Whereas an un-grouped Graph View might have this subscription:
```

hjoliver:
A sanity check from @dwsutherland would be good, before merge.

docs/proposal-subscriptions.md
oliver-sanders: This was an intentional pun!

May need to change this anyway to help separate Cylc "Views" (e.g. tree, dot, graph) from Vue "views" (i.e. application routes / page model thinggies).

hjoliver: Oh, was it ... he heh :grin:

hjoliver:
(@dwsutherland is away on leave until next week)

docs/proposal-subscriptions.md
```diff
@@ -0,0 +1,280 @@
...
+workflows(ids: [<id>]) {
+  taskProxies {
+    id: <id>,
+    status: "running"
+  }
+}
+      </pre>
+    </td>
+  </tr>
+</table>
```
dwsutherland: Because deltas are actually flat (only one topic at a time, i.e. `workflow` OR `taskProxy`), then we would need to drive the subscription off the data-store and not filter out the fields parent to the delta (even though the parent is not in the delta).. Otherwise we would need to enforce flat subscriptions, and probably construct new delta graphql object types to send the context of the delta (like protobuf deltas).

Consider:
```
workflows(ids: []) {
  taskProxies {
    id: ,
    status: "running"
    jobs {
      id: ,
      status: "running"
    }
  }
}
```
If only `jobs` topic is sent by the WFS, then the delta may have no information about `taskProxy` or `workflow`.. The delta itself is received at the UIS with the workflow and topic context though...

hjoliver:
Ping @oliver-sanders - can you address David's comments here so we can merge this?

docs/proposal-subscriptions.md
```diff
@@ -0,0 +1,280 @@
...
+    * Enables the UIServer to dynamically adjust its subscriptions to the WS so as to keep load on the workflow to a minimum.
+
+By reducing the about of data we send to the UI we reduce the load on the browser enabling us to do more with the UI and to make it more responsive.
+
+## The Proposal
+
+1. Individual Vues register and update their subscriptions with a service in the UI *(done)*.
+2. This service "merges" subscriptions together then registers the result with the UIS *(done)*.
+3. The UIS parses and stores these subscriptions.
+4. When deltas arrive from the WS, the UIS loops through active subscriptions to see if there is an overlap.
```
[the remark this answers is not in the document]
hjoliver: (now we're using "scheduler")

docs/proposal-subscriptions.md
oliver-sanders: Is this not what the intersection idea solves?

Using your example:

*Subscription:*

```graphql
workflows(ids: ["u|w"]) {
  taskProxies {
    id
    status
    jobs {
      id
      status
      submitNumber
      batchSystem
    }
  }
}
```

*Delta:*

```json
{
  "workflows": {
    "id": "u|w",
    "taskProxies": [
      {
        "id": "u|w|c|t",
        "jobs": {
          "id": "u|w|c|t|j",
          "status": "running"
        }
      }
    ]
  }
}
```

*Intersection:*

```json
{
  "workflows": {
    "id": "u|w",
    "taskProxies": [
      {
        "id": "u|w|c|t",
        "jobs": {
          "id": "u|w|c|t|j",
          "status": "running"
        }
      }
    ]
  }
}
```

docs/proposal-subscriptions.md
```diff
@@ -0,0 +1,280 @@
...
+
+```
+taskProxies {
+  id: <id>,
+  status: "running"
+  jobs {
+    id: <id>,
+    status: "running"
+  }
+}
```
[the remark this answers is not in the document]
oliver-sanders: These are illustrative examples designed to show how the single data model is synced between the three data stores.

The Scheduler-UIS part is done with Protobuf rather than GraphQL but the data is the same so I'ved represented it using GraphQL here to make things easier to understand.

oliver-sanders:
> Happy with the proposal, however:
>
> As a consequence of the proposed nested GraphQL subscriptions

This proposal should be independent of how the subscriptions are issued, it's equally valid if we use flat or nested subscriptions. We could use nested subscriptions with flat deltas or flat subscriptions with flat deltas.

I think we're all pretty set on flatland though - https://www.xkcd.com/721/

docs/proposal-subscriptions.md
dwsutherland: My point is; if your delta is flat and all you receive at the UIS is:
```
        "jobs": {
          "id": "u|w|c|t|j",
          "status": "running"
        }
```
Then you don't/may-not have the information in the delta to resolve the other bits:
```
{
  "workflows": {
    "id": "u|w",
    "taskProxies": [
      {
        "id": "u|w|c|t",
.
.
.
```
(meaning you get nothing, not even jobs)

So you will need to use the data-store to fill in the missing bits before sending it on to the WUI.. This isn't hard to do, and we could make it so if the WUI subscribes to parent bits that aren't in the driving deltas then it (optionally) only receives the skeleton (IDs) of the parent objects..

The branches I'm working on clump related deltas together (which will help).. anyway these are finer points to resolve along the way.

docs/proposal-subscriptions.md
dwsutherland: Yes all figured out, PRs tomorrow 🤞

docs/proposal-subscriptions.md
oliver-sanders: > So you will need to use the data-store to fill in the missing bits before sending it on to the WUI.

Not for subscriptions because the required data will already have been synced, should be no need to go to the data store.

The delta between the Scheduler and UIS should be exactly the same as the delta between the UIS and WUI.

dwsutherland: > Not for subscriptions because the required data will already have been synced

which means your subscription should be flat,  which is my contention.

dwsutherland: > Not for subscriptions because the required data will already have been synced

which means your subscription should be flat,  which is my contention.

PR #77, opened 2020-01-29
proposal: mutl-user functionaltiy + cylc flow

oliver-sanders:
I think this is a decision we have already unconsciously made but one we have no record of?

Push multi-user functionality out of Cylc Flow and into the UI Server.

Formally retire the *requirement* for port scanning, anomalous authentication, etc.

### DIFF
--- a/docs/proposal-multi-user-approach.rst
+++ b/docs/proposal-multi-user-approach.rst
@@ -6,7 +6,7 @@ The Background
 
 Pre Cylc8 there was some level of multi-user functionality offered by Cylc.
 
-We had the "anomalous" user, the ability to scan other users suites and HTTP(s)
+We had the "anonymous" user, the ability to scan other users suites and HTTP(s)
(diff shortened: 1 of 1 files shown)


hjoliver:
>  anomalous authentication

anonymous?

docs/proposal-multi-user-approach.rst
```diff
@@ -0,0 +1,25 @@
+Multi-User Functionality & Cylc Flow
+====================================
+
+The Background
+--------------
+
+Pre Cylc8 there was some level of multi-user functionality offered by Cylc.
+
+We had the "anomalous" user, the ability to scan other users suites and HTTP(s)
```
hjoliver:
```suggestion
We had the "anonymous" user, the ability to scan other users suites and HTTP(s)
```

(Although I really like "anomalous user" :grin: !)

PR #78, opened 2020-01-30
Proposal for changes to the config files

dpmatthews:
Addresses cylc/cylc-flow#3422.
Intended for discussion at the workshop.

I've attempted to go through all the previous discussion and list everything that was proposed.
Hopefully we can agree which changes are worthwhile and then implement them incrementally.
Many of the changes appeared to have general support in the previous discussions which I've referenced so hopefully we won't need to debate them all in detail.

Reviewers: please just check that nothing obvious is missing or wrong.


docs/proposal-config-changes.md
```diff
@@ -0,0 +1,731 @@
...
+       * Proposal: `config.rc`?
+     * Read from `/etc/cylc/flow/8.0a1/flow.rc` and
+       `~/.cylc/cylc/flow/8.0a1/flow.rc` unless `$CYLC_CONF_PATH` set in which
+       case it is read only from this path.
+     * We need the ability to override where the site config file is read from,
+       e.g. by setting `$CYLC_SITE_CONF_PATH`?
+     * Should we include the version in the path? Users should not have to
+       create a new config file every time we deploy a new version. Perhaps
+       major version only (with backwards compatibility assumed for minor
+       versions)?
```
hjoliver: This would also require forward compatibility - which can be achieved by ignoring rather than failing unkown items. pros and cons...  (you do this with Rose configs already I think?)

dpmatthews: Not necessarily. You just can't use new features whilst still using old versions. We've lived with this until now.

docs/index.md
```diff
@@ -29,6 +29,8 @@
   - Specification for new CLI - [WIP]
 
 - [task state, output, and trigger name rationalization](proposal-state-names.md)
+
+- [config file changes](proposal-config-changes.md)
```
wxtim:
```suggestion
- [config file changes](proposal-config-changes.md) - A proposal for discussion at Cylc Workshop in Febuary 2020
```

docs/proposal-config-changes.md
```diff
@@ -0,0 +1,731 @@
...
+
+## General points
+
+### Terminology
+   * We should remove all references to `suite` in the settings (partly because
+     we're trying stop referring to "suites" and partly because some affected
+     settings are actually referring to the workflow server which is unclear).
+     * Note: "suite" is referred to through-out the code, tests and
+       documentation.
+   * Do we want to use `server` as shorthand for `workflow server`?
```
wxtim: > Explicit is better than implicit.
Tim Peters

docs/proposal-config-changes.md
```diff
@@ -0,0 +1,731 @@
...
+       just leaves things from 7.8.x.
+
+### Config file naming
+
+See [cylc-flow #3170](https://github.com/cylc/cylc-flow/issues/3170).
+   * Suite: Currently `suite.rc` (specified in
+     [suite_files.py](https://github.com/cylc/cylc-flow/blob/master/cylc/flow/suite_files.py#L123)).
+     Since we're trying to move away from referring to "suites" we should change
+     this (with backwards compatibility).
+     * Proposal: `flow.rc` or `cylc-flow.rc`
```
wxtim: Other suggestions pop into my head. Will bring these up at meeting.

docs/proposal-config-changes.md
```diff
@@ -0,0 +1,731 @@
...
+   * Global
+     * Currently `flow.rc` (specified in
+       [globalcfg.py](https://github.com/cylc/cylc-flow/blob/master/cylc/flow/cfgspec/globalcfg.py#L325)).
+       * Proposal: `config.rc`?
+     * Read from `/etc/cylc/flow/8.0a1/flow.rc` and
+       `~/.cylc/cylc/flow/8.0a1/flow.rc` unless `$CYLC_CONF_PATH` set in which
+       case it is read only from this path.
+     * We need the ability to override where the site config file is read from,
```
wxtim: This sounds like it would make testing easier - I like the idea of being able to set both the site and user conf paths separately.

docs/proposal-config-changes.md
hjoliver: (I thought you were suggesting not have version-specific config files, in which case old cylc versions have to be able to handle newer config files ... which implies old cylc versions have to ignore - not fail - new or changed config items, doesn't it?)

oliver-sanders: In Cylc7 we support version-specific OR non-version specific I believe.

I think going version specific makes the upgrade process unnecessarily difficult, particularly for independent installations.

We might be able to reduce the pain by being less specific e.g:

```
~/
  `- /.cylc/
     `- flow/
        |- 8/
        |  `- flow.rc
        |- 8.0/
        |  `- flow.rc
        `- 8.0.1/
           `- flow.rc
```

hjoliver: > In Cylc7 we support version-specific OR non-version specific I believe.

Non version-specific?  Not really, because we fail on encountering unknown config items.  So it only works if new versions don't add new items.

oliver-sanders: > **Non version-specific? Not really**

We do support version-specific config files at Cylc7:

https://github.com/cylc/cylc-flow/blob/128214a0f841d89694c76da67e458cd41e3363e7/lib/cylc/cfgspec/globalcfg.py#L376

hjoliver: He heh, no - I was saying the we don't really support **non version-specific** config files.

We do support the non-specific file location, but if new cylc versions add new config items (and the new items are used in the file) it will break old versions of cylc that try to read the file.

docs/proposal-config-changes.md
hjoliver: (We've now abandoned the term "workflow server")

docs/index.md
hjoliver: No need to add this - all things can be discussed anyway :grin:

docs/proposal-config-changes.md
hjoliver: (updated post merge)

docs/proposal-config-changes.md
hjoliver: (Added @oliver-sanders comment post-merge) (the cylc7 discussion isn't relevant)

oliver-sanders:
FYI: The Scheduler already "unifies" the workflow and global configurations for all settings in the `[cylc]` section, or at least it looks for the configuration in both.

https://github.com/cylc/cylc-flow/blob/166a51859764e54a7a1cbde8b072b0357ca9670e/cylc/flow/scheduler.py#L2087-L2099

PR #80, opened 2020-02-04
Component naming proposal.

hjoliver:
Let's see if we can finally lay this long-running discussion to rest.

Note this proposes one repository name change too (cylc-webui) for reasons explained in the document. Somewhat painful, but there will never be a better time than now.

@cylc/core - if anyone disagress with one or more of the marked (:boom:) changes at first glance, please read the relevant "Explanatory Comments" before deciding ... or else approve for merge.

Assigning everyone for this - if you don't have a strong opinion on it, just approve :grin: 

(Read with markdown rendered, via "View File" )

### DIFF
--- a/docs/proposal-component-names.md
+++ b/docs/proposal-component-names.md
@@ -38,7 +38,7 @@ Cylc Flow Instance
 
 Cylc config files: `cylc-flow.conf`, `cylc-flow-global.conf`
 - (we're dropping the term "suite" - see above)
-- the `.rc` post-fix is archaic, rarely used today 
+- the `.rc` suffix is archaic, rarely used today 
 - the old `global.rc` is too generic
 - :question: `flow.conf` is an option
   - (but `cylc-flow.conf` is clearer without being too verbose)
--- a/docs/proposal-component-names.md
+++ b/docs/proposal-component-names.md
@@ -44,12 +44,12 @@ Cylc config files: `cylc-flow.conf`, `cylc-flow-global.conf`
   - (but `cylc-flow.conf` is clearer without being too verbose)
 
 Cylc Hub
-- A vanilla (so far) instance of JupyterHub, BUT
-  - Configured to show Cylc branding and launch the Cylc UI Server
-- To avoid confusion with other JupyterHub on the system we need to call it
-  "Cylc Hub", but we should be clear that it is just an instance of JupyterHub
-- :question: do we need to launch it with a special command: `cylc hub`? (which
-  automatically picks up the Cylc-specific `jupyterhub_config.py`?)
+- A vanilla (so far) JupyterHub instance, configured to show Cylc branding and
+  launch the Cylc UI Server
+- To avoid confusion, and in case we ever need to go non-vanilla, we will call
+  it "Cylc Hub" - but we should be clear that it is actually just JupyterHub
+- :question: do we need to launch it with a special Cylc command that 
+  automatically picks up the Cylc-specific `jupyterhub_config.py`?
 
 Cylc UI Server
 - The word "Server" is not so confusing here (as it could be for the scheduler
--- a/docs/proposal-component-names.md
+++ b/docs/proposal-component-names.md
@@ -1,0 +1,62 @@
+# Proposal (Feb 2020)
+
+## Cylc 8 System Component Naming
+
+Cylc is now a complex multi-component system. It is important that we settle on
+consistent, unambiguous, and where possible standard, component names so that
+users and developers can communicate without confusion. (This was not easy even
+before Cylc 8!).
+
+## Proposed Final Names
+
+Firstly, we have already agreed to drop **suite** for **workflow** throughout
+("suite" is non-standard outside of the NWP community).
+
+Secondly, we should name the various system components as follows (changes marked with :boom:):
+
+| Component Name               | Code Repository   | Config File                                          | Description |
+| --                           |----               | ---                                                  |---          |
+|:boom: **Cylc Flow Instance** | cylc-flow         | :boom: `cylc-flow.conf` and `cylc-flow-global.conf`  | an instance of the cylc-flow program, to manage a single workflow | 
+|**Cylc Hub**                  | (Jupyter Hub)     | :boom: `cylc-hub.conf`                               | the single point of access that users log in to |
+|**Cylc UI Server**            | cylc-uiserver     | N/A? (or :boom: `cylc-uiserver.conf`)                | an instance of the cylc-uiserver program, that serves the Cylc Web UI and CLI |
+|:boom: **Cylc Web UI**        | :boom: cylc-webui | N/A?                                                 | an instance of the Cylc Web UI, served to your browser by the Cylc UI Server |
+
+
+## Explanatory Comments
+
+Cylc Flow Instance
+- "workflow server" is ambiguous - it can refer to the host server that the workflows run on
+- "workflow server program" is clearer, but too much of a mouthful.
+- "workflow daemon" is not strictly correct (`--no-detach`) and is very
+  Unix-specific (should we ever port to another OS)
+- "workflow service" - a cylc-flow instance really stretches the definition of a "service"
+- "cylc-flow instance" is strictly correct and unambiguous, reasonably
+  intuitive (a running workflow instance), and its function in the context of the wider system
+  can be described more verbosely if needed (see the Description in the table above)
+- :question: "Flow Instance" is an option
+  - (but "Cylc Flow Instance" is clearer without being too verbose)
+
+Cylc config files: `cylc-flow.conf`, `cylc-flow-global.conf`
+- (we're dropping the term "suite" - see above)
+- the `.rc` post-fix is archaic, rarely used today 
+- the old `global.rc` is too generic
+- :question: `flow.conf` is an option
+  - (but `cylc-flow.conf` is clearer without being too verbose)
+
+Cylc Hub
+- A vanilla (so far) instance of JupyterHub, BUT
+  - Configured to show Cylc branding and launch the Cylc UI Server
+- To avoid confusion with any standard JupyterHub on the system
+  - we should call it "Cylc Hub", and:
+  - it should be launched with a special command: `cylc hub`
+  - it should have an obvious Cylc-specific config file: `cylc-hub.conf`
+
+Cylc UI Server
+- The word "Server" here potentially suffers from the "host server" ambiguity too, but:
+  - it is a more appropriate term here in the sense that the UI Server exists entirely to "serve the UI"
+  - (for cylc-flow we have to distinguish between the instance and the
+    codebase, which includes the CLI etc. as well as the "server program")
+
+Cylc Web UI
+- The "Cylc UI" is not just the web interface, it's also the CLI
+- No one uses the term "GUI" for web interfaces
--- a/docs/proposal-component-names.md
+++ b/docs/proposal-component-names.md
@@ -9,53 +9,51 @@ before Cylc 8!).
 
 ## Proposed Final Names
 
-Firstly, we have already agreed to drop **suite** for **workflow** throughout
-("suite" is non-standard outside of the NWP community).
-
-Secondly, we should name the various system components as follows (changes marked with :boom:):
+Note we have already agreed to drop **suite** for **workflow** ("suite" is
+non-standard outside of the NWP community).
 
 | Component Name               | Code Repository   | Config File                                          | Description |
 | --                           |----               | ---                                                  |---          |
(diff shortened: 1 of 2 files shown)


hjoliver:
On the proposed config file names, I think we were aiming for `(cylc)-flow.rc`.  However, see my arguments against the `.rc` suffix, and if we're change the main file name already, changing the suffix as well is not really an additional burden on users.

hjoliver:
> Only minor nit pick I have is that I think JupyterHub would be fine...

That's the bit I feel least sure of, for the same reasons you give. 

But the problem is for us:
- some Cylc users will not be familiar with JupyterHub, or won't have it on the system (apart from Cylc)
- even if JupyterHub is also being used for Jupyter Notebooks on the same system, we have to pick up a non-default config file (from a Cylc-specific location) ... so a cylc-specific wrapper (`cylc hub` command) would hide that detail from users/admins.

On the other hand,
- I suppose it's not that hard to document the cylc-specific hub config file or command-line options
- it does feel a bit like stealing, to completely hid the fact that we're using JupyterHub out of the box (I don't think I'd like that if I was on their team...) - maybe we should not do it, just for this reason!

Let's see what others think...

dpmatthews:
I think we need to be able to refer to cylc hub for the reasons given. It will (hopefully) just be a cylc branded/configured jupterhub and we can make that very obvious. I'm not yet sure if we'll need a "cylc hub" command or a specific config file.

I'm not an immediate fan of "cylc flow instance". We're proposing a new section in the config file for configuring the server - what would than become? cylc message "communicates back to the instance" - that doesn't sound right? Do we write to the "instance log file"? I understand the issue with "server" and don't really like it although if we're happy to stick with server for the UIS then I'd have thought it would be OK for the WFS as well. I've suggested "controller" before. "service" is still an option (not perfect but nothing is) but maybe use it for both UIS & WFS?

matthewrmshin:
Chipping in, since I got bombarded by notifications of this pull request for some reason...

You may want to consider the phrase *ui-backend* and *ui-frontend*, as opposed to *ui-server* and *web-ui*.

hjoliver:
@dpmatthews 
> I'm not an immediate fan of "cylc flow instance". We're proposing a new section in the config file for ...

Point taken, but in a config file a little more verbosity doesn't matter, to get clarity.  We mainly need concise "names" in order to talk about the components easily (t

hjoliver:
(Aarggh, accidental keyboard disaster, going back to continue the previous comment...)

hjoliver:
@matthewrmshin - nice to hear from you :+1: 

> You may want to consider the phrase ui-backend and ui-frontend, as opposed to ui-server and web-ui.

That's a new idea (as far as I'm aware).  Personally I think that the name of the WFS is the harder thing to settle on, and "Web UI" and "UI Server" are relatively sensible, correct, and intuitive terms. But if others disagree with me, we can consider your suggestion.

hjoliver:
(Not entirely surprising, but it seems we might not be able to settle this without another discussion about it in the workshop).

docs/proposal-component-names.md
```diff
@@ -0,0 +1,62 @@
...
+- "workflow daemon" is not strictly correct (`--no-detach`) and is very
+  Unix-specific (should we ever port to another OS)
+- "workflow service" - a cylc-flow instance really stretches the definition of a "service"
+- "cylc-flow instance" is strictly correct and unambiguous, reasonably
+  intuitive (a running workflow instance), and its function in the context of the wider system
+  can be described more verbosely if needed (see the Description in the table above)
+- :question: "Flow Instance" is an option
+  - (but "Cylc Flow Instance" is clearer without being too verbose)
+
+Cylc config files: `cylc-flow.conf`, `cylc-flow-global.conf`
```
wxtim: can we modify this so that a `cylc-flow-user.conf` and `cylc-flow-site.conf` are used to create the global config object? 

Also is the "flow" necessary (or even a good idea) in the global config names?

wxtim:
My feelings are not as strong as @dpmatthews on this, but I do think that we can do better than instance - controller might be preferable or even `a cylc workflow`.

My feelings are not strong enough on this matter to debate it for too long, so I've approved having inserted my two-pennyworth.

docs/proposal-component-names.md
hjoliver: > can we modify this so that a cylc-flow-user.conf and cylc-flow-site.conf are used to create the global config object?

I think that's only necessary if there are real differences between the user and site global config (i.e. are we proposing site global config items that users cannot override in their global config file? ... if not, making the `-user.conf` and `-site.conf` distinction seems a bit superfluous to me.

> Also is the "flow" necessary (or even a good idea) in the global config names?

 My reasoning here was that the global and workflow config files both configure the same program, i.e. the "cylc flow instance" aka WFS etc.  So the config file name contains the name of the program that it configures.  I guess it's not strictly necessary, but if it is not overly verbose or ungainly, then it is (IMO) nicely self-documenting in that respect.

hjoliver:
@wxtim - 

>  controller might be preferable or even "a cylc workflow"

We have considered "controller" before, but the problem is it can be confused with the UI that the user uses to "control" the running workflow.

Also "cylc workflow" already refers primarily to the abstract workflow - i.e. the graph of task dependencies.  "cylc-flow instance" has the virtue (IMO) of explicitly suggesting an (instantiated) "instance" of the "cylc-flow" program (not an instance of the workflow as such, but of the program that manages the workflow).

hjoliver:
Summary so far:

- we should call the hub component "Cylc Hub" to avoid any confusion with normal use of J-Hub (and in case we ever have to branch off of the vanilla J-Hub). But whether or not we should launch it with a special cylc command remains to be decided (that's fine, it's a lower level concern)

- config file names: I think I have addressed @wxtim 's concerns above, and he approved as-is so we'll provisionally stick with the names proposed with final decision in the workshop

- component names (the primary concern here): the main sticking point is the name of the "suite daemon" (in the old parlance). But @dpmatthews may have just solved the problem for us by suggesting (offline) **Cylc Scheduler**. I think that avoids the issues with service/server *and* it is both semantically correct and more intuitive than "instance".  And it rolls off the tongue better too.  I can't think why we have not considered this before!  

Main Components, from front to back:

- Cylc Hub
- Cylc Web UI
- Cylc UI Server
- Cylc Scheduler

 Any objections?

hjoliver:
(Doh, looks like I "assigned" everyone as owners not reviewer ... will un-assign now...)

hjoliver:
After chat with @oliver-sanders - 

- the term "scheduler" nicely describes what the program actually does; this might help clarify things for users who sometimes don't get the difference between a "suite" and a "job"

- the program is actually largely an instance of the `Scheduler` class from `cylc/flow/scheduler.py` - duh!!!  (Why didn't we think of this before :thinking: )

- (Oliver now agrees).

I will update the proposal (and indicate that the config file names are still somewhat provisional until we're done with that section of the workshop) ... and merge this PR

docs/proposal-component-names.md
```diff
@@ -0,0 +1,60 @@
...
+before Cylc 8!).
+
+## Proposed Final Names
+
+Note we have already agreed to drop **suite** for **workflow** ("suite" is
+non-standard outside of the NWP community).
+
+| Component Name               | Code Repository   | Config File                                          | Description |
+| --                           |----               | ---                                                  |---          |
+|:boom: **Cylc Scheduler**     | cylc-flow         | :boom: `cylc-flow.conf` and `cylc-flow-global.conf`  | an instance of the cylc-flow scheduler program, to manage a single workflow |
```
oliver-sanders:
```diff
- `cylc-flow.conf` and `cylc-flow-global.conf`
+ `flow.rc`
```

docs/proposal-component-names.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Proposed Final Names
+
+Note we have already agreed to drop **suite** for **workflow** ("suite" is
+non-standard outside of the NWP community).
+
+| Component Name               | Code Repository   | Config File                                          | Description |
+| --                           |----               | ---                                                  |---          |
+|:boom: **Cylc Scheduler**     | cylc-flow         | :boom: `cylc-flow.conf` and `cylc-flow-global.conf`  | an instance of the cylc-flow scheduler program, to manage a single workflow | 
+|**Cylc Hub**                  | (Jupyter Hub)     | :boom: `<cylc-location>/jupyterhub_config.py`               | the single point of access that users log in to |
```
oliver-sanders:
```diff
- `<cylc-location>/jupyterhub_config.py`
+ hub.py
```

docs/proposal-component-names.md
```diff
@@ -0,0 +1,60 @@
...
+|**Cylc UI Server**            | cylc-uiserver     | N/A? (or :boom: `cylc-uiserver.conf`)                | an instance of the cylc-uiserver program, that serves the Cylc Web UI and CLI |
```
oliver-sanders:
```diff
- `cylc-uiserver.conf`
+ `uiserver.py` ?
```

docs/proposal-component-names.md
```diff
@@ -0,0 +1,60 @@
...
+|:boom: **Cylc Web UI**        | :boom: cylc-webui | N/A?                                                 | an instance of the Cylc Web UI, served to your browser by the Cylc UI Server |
```
oliver-sanders: Note: filename / storage determined by whether we can store user preferences using Jupyterhub.

docs/proposal-component-names.md
hjoliver: Re `.rc` - see my arguments against that in the "Explanatory Comments" bit of the document.

Are  you suggesting `flow.[ext]` for BOTH workflow-specific and global config, even though we are no longer unifying the files?  Not sure about that, as we won't be able to tell which is which by merely inspecting the filename (they might not always be stored in the their correct location).

I could go for `flow.conf` (global) and `cylc-flow.conf` (workflow) IF the site-global location includes `cylc` in the directory hierachy like the user-global location does - but I don't think we've talked about that yet?

docs/proposal-component-names.md
hjoliver: I'll note this in the proposal as an option (I did originally suggest `cylc-hub.py`) but I think leaving it as  `jupyterhub_config.py` in the cylc-specific location would be fine - no reason to hide the fact that "Cylc Hub" is in fact just JupyterHub, and users won't normally have to deal directly with hub config.

docs/proposal-component-names.md
hjoliver: Again, only if we "cylc" in a directory hierachy that the file is required to reside in.

docs/proposal-component-names.md
oliver-sanders: > see my arguments against that in the "Explanatory Comments" bit of the document.

Just read. I think this change is probably all pain, no gain.

Users are going to hate us for renaming `suite.rc` in the first place.

> Are you suggesting flow.[ext] for BOTH workflow-specific and global config

Meh.

> I could go for flow.conf (global) and cylc-flow.conf (workflow) 

At the moment we are using `flow.rc` for both, I don't really mind if we call them the same thing or use different names.

I don't like `cylc-flow.conf` for workflow definitions, the `cylc-` is superfluous and more to type.

Why not `~/.cylc/flow/global.conf` (global) and `flow.conf` (workflow)?

"We refer to workflows as flows which are defined a in a `flow.conf` file"

docs/proposal-component-names.md
oliver-sanders: We have control of this file / search path so we can call it what we want but fair point.

docs/proposal-component-names.md
hjoliver: > Users are going to hate us for renaming suite.rc in the first place.

Huh, well, too bad - we can't standardize our terminology for the longer-term good without getting rid of the word "suite", and we agreed to do that long ago. So users already have to change their workflow configuration file names.  So why not make them change the last two  characters *in a filename that they already have to change the rest of* - to a suffix that is also more standard and clearly indicates a config file (no one knows what ".rc" means any more).

hjoliver: > I don't like `cylc-flow.conf` for workflow definitions, the `cylc-` is superfluous and more to type.
> Why not ~/.cylc/flow/global.conf (global) and flow.conf (workflow)?

The `cylc-` is only five characters, you don't have to type it very often, and it is only superfluous if the directory hierachy contains "cylc" - which it does for the user global file, but not for the workflow file, and it may or may not for the site global file.
  
Plain `global.conf` is extremely generic - if you find one of those lying around god only knows what it is for. 

For the workflow config `flow.conf` also seems a bit too generic to me, given that half the workflow engines on the planet are likely to call their workflows "flows" too.  Why not make it completely obvious what the file is for with only five extra characters? `cylc-flow.conf`

docs/proposal-component-names.md
oliver-sanders: Which is the plan?

hjoliver: Like the cylc-flow global file, not decided yet as far as I'm aware.

PR #81, opened 2020-02-09
Minor update to getting started with the UI

wxtim:
@oliver-sanders has advised me that developers probably want to install all optional dependencies - so I have updated my destructions.

### DIFF
--- a/docs/howto/cylc-ui.md
+++ b/docs/howto/cylc-ui.md
@@ -19,7 +19,8 @@
    [cylc/cylc-flow](https://github.com/cylc/cylc-flow/) and
    [cylc/cylc-uiserver](https://github.com/cylc/cylc-uiserver/):
   ```bash
-  pip install -e "${DEVDIR}/cylc-flow" -e "${DEVDIR}/cylc-uiserver"
+  pip install -e "${DEVDIR}/cylc-flow"[all]
+  pip install -e "${DEVDIR}/cylc-uiserver"[all]
   ```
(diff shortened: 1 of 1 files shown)


docs/howto/cylc-ui.md
```diff
@@ -19,7 +19,8 @@
    [cylc/cylc-flow](https://github.com/cylc/cylc-flow/) and
    [cylc/cylc-uiserver](https://github.com/cylc/cylc-uiserver/):
   ```bash
-  pip install -e "${DEVDIR}/cylc-flow" -e "${DEVDIR}/cylc-uiserver"
+  pip install -e "${DEVDIR}/cylc-flow"[all]
+  pip install -e "${DEVDIR}/cylc-uiserver"[all]
```
kinow: I think the [all] needs to go inside the quotes? Does it work this way too?

hjoliver: That will work fine, but it does look a bit funny.  Strictly only `${DEVDIR}` needs quoting in case of spaces in the path.  So I'd go for just the variable in quotes, or the whole string in quotes.

kinow: Oh, yup. The variable expansion in shell was clear. But it's because I only had to use quotes with this command as `pip install ."[all]"` or `pip install ".[all]"`, [e.g.](https://github.com/cylc/cylc-flow/blob/166a51859764e54a7a1cbde8b072b0357ca9670e/.travis/install.sh#L38)... though I don't remember why. Not important, marking as resolved. Sorry for the noise!

PR #84, opened 2020-02-26
Added Howto: Inspect a suite database

wxtim:
A couple of months ago I was wanting to poke the suite database for the database upgrader ticket. I have very little experience with SQL and the suite data was a mystery to me. I wrote this to capture some of the advice of @oliver-sanders on interrogating the suite database and to document what I learned for the future. 
Reviewers should consider:
- Is this the right place for this document? I think it is because interrogating the suite database is a developer/support task, so shouldn't be in the main docs.
- Is this sufficiently sanitized? Is there anything I've left in the document which gives away information about me.
- Are the instructions good enough to make it work? Do I need to explain how to ensure that you have sqlite (I have mine in my conda env)?


hjoliver:
@wxtim - thanks, this probably is useful - although I can't tell for sure until I've reminded myself how to use Jupyter Notebooks (somewhat ironic, give our new Hub, he heh).  In the meantime @kinow might want to take a quick look.

> Is this the right place for this document? I think it is because interrogating the suite database is a developer/support task, so shouldn't be in the main docs.

Here will do for now, we can always move it later.  Interrogating the DB isn't necessarily just for developers (users might want to analyze their suite history) ... but guess a Python notebook wouldn't fit in the user guide anyway.

kinow:
> @wxtim - thanks, this probably is useful - although I can't tell for sure until I've reminded myself how to use Jupyter Notebooks (somewhat ironic, give our new Hub, he heh). In the meantime @kinow might want to take a quick look.

I normally use nbviewer.org. GitHub UI will break while trying to render notebooks, and sometimes I don't have the dependencies locally to render/eval notebook cells.

https://nbviewer.jupyter.org/github/wxtim/cylc-admin/blob/suite-db-inspection-notebook/docs/howto/inspect_suite_database.ipynb

Clicking above you should be able to visualize the notebook @hjoliver . You can visualize any notebook on GitHub with that URL, just need to play with the values after `github/`.

The notebook looks OK to me. Notes that may be useful to other developers.

If I am in a terminal-only session, I also use `sqlite3` with normal SQL combined with `.schema`, `.tables` SQLite specific commands, but remembering the tables and relationships can be tricky. But if I have a GUI, then I normally use only [`sqlitebrowser`](https://sqlitebrowser.org/).

hjoliver:
(Yes I also use sqlitebrowser and sqlite3 CLI).

hjoliver:
>Is this sufficiently sanitized? Is there anything I've left in the document which gives away information about me.

I guess that's for you to decide @wxtim ! (only you know what info you're happy to give away).

hjoliver:
> Is this sufficiently sanitized? 

Actually, since you asked, the notebook does contain a bunch of specific hostnames, and your user name.

docs/howto/inspect_suite_database.ipynb
```diff
@@ -0,0 +1,473 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 105,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Set the parameters of the suite whose database you want to investigate\n",
+    "USER = \"jbloggs\"\n",
+    "SUITENAME = \"simplest\"\n",
+    "SUITE_FILE_GLOB_PATTERN = f\"/home/*/{USER}/cylc-run/{SUITENAME}/\"\n",
```
hjoliver: This doesn't work for the usual `/home/$USER/` home dir structure.  
Maybe better to have a separate home dir variable, so it's more obvious that that needs customizating, and then automatically add the standard cylc-run path after that?

docs/howto/inspect_suite_database.ipynb
```diff
@@ -0,0 +1,473 @@
...
+    "\n",
+    "# Deliberately left blank before sharing\n",
+    "SQLLITEPATH = \"/home/h02/tpilling/miniconda3/bin/sqlite3\""
```
hjoliver: Exposed user name and home dir path?

docs/howto/inspect_suite_database.ipynb
```diff
@@ -0,0 +1,473 @@
...
+      ]
+     },
+     "execution_count": 103,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "from io import StringIO\n",
+    "pd.read_csv(\n",
```
hjoliver:
```python
NameError                                 Traceback (most recent call last)
<ipython-input-8-513363772c68> in <module>
      1 from io import StringIO
----> 2 pd.read_csv(
      3     StringIO(M.stdout),
      4     sep="|",
      5     index_col=False,

NameError: name 'pd' is not defined
```

docs/howto/inspect_suite_database.ipynb
hjoliver: (Note I've deliberately not highlighted the exposed internal host names)

docs/howto/inspect_suite_database.ipynb
wxtim: Yes, so I've changed this to using
```python
os.getenv('HOME')
```

docs/howto/inspect_suite_database.ipynb
wxtim: Yep, there needs to be a pandas import somewhere. I'm guessing that it worked for me because I'd loaded pandas in that session and it remained available, even when I deleted the code which had imported it. Jupyter Notebooks can be a pain on that front.

wxtim:
@hjoliver, I think I've sorted out those items. To run it:
```bash
conda activate environment-with-jupyter-pandas-stack
jupyter notebook /path/to/notebook.ipynb
```

PR #86, opened 2020-04-07
Platforms proposal: add details to the rose suite-run functionality

dpmatthews:
This covers:
* Installation of suites on platforms
* Symlinking of directories to different locations

### DIFF
--- a/docs/proposal-platforms.md
+++ b/docs/proposal-platforms.md
@@ -50,7 +50,7 @@ and partially address
 
 ```ini
 [platforms]
-    [[desktop\d\d|laptop\d\d]]
+    [[desktop\d\d,laptop\d\d]]
         # hosts = platform name (default)
         # Note: "desktop01" and "desktop02" are both valid and distinct platforms
     [[sugar]]
@@ -235,11 +235,11 @@ suite-run functionality:
      before the first set of jobs is submitted to each platform after suite
      start-up or on suite reload. This system needs to be extended to other
      items of the suite.
-   * We will modify this to work on a target / filesystem basis rather than a
+   * We will modify this to work on an install target basis rather than a
      platform basis.
      * Configure the id of the target used on a particular platform
        (`install target`: defaults to the platform name).
-     * If multiple platforms share the same root filesystems then use a common
+     * If multiple platforms share the same filesystems then use a common
        `install target`.
      * Note that, when we add we support for separate client keys, these will
        now be per `install target` rather than per platform.
@@ -248,7 +248,7 @@ suite-run functionality:
        functionality. Instead we should check for the existence of a client key
        and fail the remote-init if one is found for a different `install target`
        (different targets used within the same workflow should not share the
-       same workflow directory).
+       same run directory).
    * Currently, the files which need to be installed are transferred as a tar
      file via stdin. This has the advantage that the remote init can be done as
      a single ssh. We will change the installation to use rsync which will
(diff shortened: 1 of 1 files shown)


docs/proposal-platforms.md
```diff
@@ -42,15 +42,15 @@ and partially address
...
+  * Unclear why we need this given that Jinja2 variables can be used?
   * Using `$ROSE_ORIG_HOST` relies on Rose specifying extra options to set this
     in `[cylc][environment]` which we should avoid if possible.
 
 ## Example platform configurations
 
 ```ini
 [platforms]
-    [[desktop\d\d|laptop\d\d]]
+    [[desktop\d\d,laptop\d\d]]
```
dpmatthews, line 53: Changed to use 2 separate patterns rather than a single one.
I assume this is already handled by parsec so should just work?
It will still be possible to use a single pattern as before if you prefer (since you can use any regular expression recognised by re.fullmatch).

docs/proposal-platforms.md
```diff
@@ -235,31 +235,183 @@ suite-run functionality:
...
+       * Note: `inherit` will imply using the same `install target`.
+     * Do we need comms method to be the same as well?
+   * The `[symlink dir]` settings are only applied when a workflow is installed
+     (or first run on a target). Therefore, changes to these settings have no
+     affect on running or restarting workflows.
+   * At the moment this proposal does not provide any way to relocate the
+     top level `$HOME/cylc-run` directory (which you can do currently via the
+     existing Cylc `run directory` setting). The assumption is that it is
+     sufficient to provide a way to move the run directory and there is no
+     reason not to use `$HOME/cylc-run`.
```
dpmatthews, line 366: Does this seem reasonable?

docs/proposal-platforms.md
```diff
@@ -235,31 +235,183 @@ suite-run functionality:
...
+       exclude any file / directories which are generated at run time or target
+       specific (`log/ share/ work/` + various files / directories in
+       `.service/`).
+       * There are some suites around which write to top level directories (not
+         sure if this is deliberate) - exclude would be very bad in these cases.
+         Therefore, include is the safer option.
+
+2. Support moving some directories to different locations with symlinks to the
+   original location.
+   * New platform settings:
```
dpmatthews, line 291: I've gone for `[symlink dir]` - alternative suggestions welcome.

docs/proposal-platforms.md
```diff
@@ -235,31 +235,183 @@ suite-run functionality:
...
+       `<work dir>/<workflow-name>/work` and a symbolic link will be
+       created from `$HOME/cylc-run/<workflow-name>/work`. If not
+       specified the workflow work directory will be created in
+       `$HOME/cylc-run/<workflow-name>/work`.
+   * What happens if you want to be able to have some workflows which use
+     different directory configurations on the same "platform"? For
+     example, you might have some workflows where you want the run
+     directory on $HOME and others on $DATADIR.
+     * For platforms other than localhost the answer is to create separate
+       platforms and then choose the appropriate platform in the workflow.
```
dpmatthews, line 333: Not as easy as the current mechanism in Rose but hopefully this is not a common requirement.

docs/proposal-platforms.md
```diff
@@ -235,31 +235,183 @@ suite-run functionality:
...
+     * We will use the `--delete` rsync option so that any files removed from
+       from installed directories also get removed from the install targets on
+       reload/restart.
+     * Does this need to be configurable? (i.e. configure additional items to
+       include)? If so, global and/or workflow level?
+       * Is there any reason why workflow designers can't just stick to the list
+         of supported directories for files that need to be installed?
+       * Are there many existing Rose suites which rely on other directories
+         being installed? A quick search implies there are quite a few suites
+         using a wide range of sub-directories such as `python`, `util`, `data`
```
wxtim, line 274: Are these reasonable? I can't help wondering if `python` and `util` should be under `lib` and `bin` respectively. `data` could feasibly go under `etc`, depending on how it is used. Can we sample some of these cases and see if it's a suite design fail or a genuine need?

dpmatthews: I agree these shouldn't be needed (see previous bullet). However, these suites already exist and currently work with rose suite-run. Hence my proposal which allows affected users to modify their global config so that they can still run these suites (until the suites can be modified to use the supported directories).

docs/proposal-platforms.md
wxtim: Would `symlink path` be more accurate?

docs/proposal-platforms.md
wxtim: Yes. 
Consider where it might cause a problem:
1. User has unrelated top level `cylc-run` dir: Just don't do this!
2. User keeps test/dev suites in a different run dir so that if something goes wrong they can delete it all (mia culpa!): Probably bad practice anyway.
3. ...

docs/proposal-platforms.md
```diff
@@ -235,31 +235,183 @@ suite-run functionality:
...
+   * New platform settings:
+     * `[symlink dir]run` (default: `none`):
+       Specifies the directory where the workflow run directories are
+       created. If specified, the workflow run directory will be created in
+       `<run dir>/<workflow-name>` and a symbolic link will be
+       created from `$HOME/cylc-run/<workflow-name>`. If not specified the
+       workflow run directory will be created in
+       `$HOME/cylc-run/<workflow-name>`. All the workflow files and the
+       `.service` directory get installed into this directory.
+     * `[symlink dir]log` (default: `none`):
```
hjoliver, line 300: :+1: thanks

PR #87, opened 2020-04-16
Add Conda Forge release instructions

kinow:
This is for Cylc 8 only. Happy to amend, maybe shorten, etc, if required. I assumed it would be better to add as much detail first, then remove what others find is in excess :+1:

### DIFF
--- a/docs/howto/create-a-release.rst
+++ b/docs/howto/create-a-release.rst
@@ -141,8 +141,9 @@ at the top of the file.
 - If you updated the version, set the build number back to 0.
 - If you did not update the version, increase the build number.
 - If the conda package contains a source URL, check that it is evaluated
-correctly using the Jinja variables, and update the ``sha256`` value
-(you can get this value from PYPI, or use some tool like ``sha256sum``).
+correctly by manually replacing the Jinja variables, and trying this out
+in a browser. Then update the ``sha256`` value (you can get this value
+from PYPI, or use some tool like ``sha256sum``).
 - Check requirements (we may have changed setup.py, or package.json, etc.
 ensure we are up to date).
 - Ensure the test commands are still valid (i.e. they can be executed
--- a/docs/howto/create-a-release.rst
+++ b/docs/howto/create-a-release.rst
@@ -180,23 +180,21 @@ and then activate it ``conda activate cylc1``. Then to build and install
 locally:
 
 .. code-block:: bash
-    # Where $CONDA_FORGE_REPOSITORY could be, for example,
-    # cylc-uiserver-feedstock.
-    cd $CONDA_FORGE_REPOSITORY
 
-    # Your package should not be listed!
-    conda list
-
-    # This will take some minutes and print useful information.
-    conda build recipe/
-
-    # The following command will install the locally created package. Before
-    # installing it will ask you to confirm. Scroll up and search the
-    # package name. The right-side column must show a location like
-    # .../anaconda3/conda-bld/linux-64::cylc-uiserver-0.1-py37_1.
-    # This confirms you are installing the local build. Here $PACKAGE_NAME
-    # could be something like cylc-uiserver.
-    conda install $PACKAGE_NAME
+   # Where $CONDA_FORGE_REPOSITORY could be, for example,
+   # cylc-uiserver-feedstock.
+   cd $CONDA_FORGE_REPOSITORY
+   # Your package should not be listed!
+   conda list
+   # This will take some minutes and print useful information.
+   conda build recipe/
+   # The following command will install the locally created package. Before
+   # installing it will ask you to confirm. Scroll up and search the
+   # package name. The right-side column must show a location like
+   # .../anaconda3/conda-bld/linux-64::cylc-uiserver-0.1-py37_1.
+   # This confirms you are installing the local build. Here $PACKAGE_NAME
+   # could be something like cylc-uiserver.
+   conda install $PACKAGE_NAME
 
 At this point you should be good to go. Test the package with commands
 such as ``which $PACKAGE_NAME``, or ``$PACKAGE_NAME --version``, etc.
--- a/docs/howto/create-a-release.rst
+++ b/docs/howto/create-a-release.rst
@@ -169,11 +169,12 @@ configured to avoid automatically upload the package. Open you ``~/.condarc``
 and check that you have something similar to:
 
 .. code-block:: yaml
-    channels:
-      - defaults
-      - conda-forge
-    ssl_verify: true
-    anaconda_upload: false
+
+   channels:
+     - defaults
+     - conda-forge
+   ssl_verify: true
+   anaconda_upload: false
 
 Now create a Conda environment for your tests, e.g.: ``conda create -n cylc1``,
 and then activate it ``conda activate cylc1``. Then to build and install
--- a/docs/howto/create-a-release.rst
+++ b/docs/howto/create-a-release.rst
@@ -102,6 +102,115 @@ Upload build to PyPI
 
 Check PyPI for your upload.
 
+Upload to Conda
+===============
+
+NOTE: the GitHub and PYPI releases are prerequisites for releasing to
+Conda Forge!
+
+The Conda Forge release process is mostly automated by bots. These bots
+are set up during the initial project bootstrap, which has been done for
+all the projects that are part of Cylc 8.
+
+As there are inter-dependencies amongst the Cylc 8 parts, you should
+know the dependency tree, and start by the leaf nodes (i.e. a module
+A may have multiple dependants, but no dependency to other modules).
+
+For example, Cylc UI Server depends on Cylc Flow. So unless you are
+releasing only Cylc UI Server, you should release Cylc Flow first.
+
+* On GitHub, navigate to the project repository on GitHub, e.g.
+
+- https://github.com/conda-forge/metomi-isodatetime-feedstock
+- https://github.com/conda-forge/cylc-uiserver-feedstock
+- https://github.com/conda-forge/cylc-ui-feedstock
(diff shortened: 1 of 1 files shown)


docs/howto/create-a-release.rst
```diff
@@ -102,6 +102,115 @@ Upload build to PyPI
...
+are set up during the initial project bootstrap, which has been done for
+all the projects that are part of Cylc 8.
+
+As there are inter-dependencies amongst the Cylc 8 parts, you should
+know the dependency tree, and start by the leaf nodes (i.e. a module
+A may have multiple dependants, but no dependency to other modules).
+
+For example, Cylc UI Server depends on Cylc Flow. So unless you are
+releasing only Cylc UI Server, you should release Cylc Flow first.
+
```
wxtim, line 121: Suggest that you add indication of current structure of dependencies:
cylc-conda-deps png
(I have the svg of that if you want)
Also suggest that you tell people where to find the info they need -Am I correct in thinking that this is `github.com/conda-forge/${REPO}/blob/master/recipe/meta.yaml` for each of the repos shown below?

docs/howto/create-a-release.rst
```diff
@@ -102,6 +102,115 @@ Upload build to PyPI
...
+This is a possible order of action:
+
+- Check open pull requests—There are bots that alert about dependencies
+(N.B. conda packages do not use PYPI for dependencies, they all use
+other conda packages as dependencies!.
+- Check if you need to update the version in the Jinja2 variable right
+at the top of the file.
+- If you updated the version, set the build number back to 0.
+- If you did not update the version, increase the build number.
+- If the conda package contains a source URL, check that it is evaluated
```
wxtim, line 143: I'm not sure I know what this step is telling me to do.
if I'm looking at:

https://github.com/conda-forge/cylc-uiserver-feedstock/blob/a7e7214c49ca420899cf3a6e7e0834cc1eff16fc/recipe/meta.yaml#L8-L10

how do I carry out this check? How do I generate the hyperlink and how do I generate the new checksum?

docs/howto/create-a-release.rst
```diff
@@ -102,6 +102,115 @@ Upload build to PyPI
...
+      - conda-forge
+    ssl_verify: true
+    anaconda_upload: false
+
+Now create a Conda environment for your tests, e.g.: ``conda create -n cylc1``,
+and then activate it ``conda activate cylc1``. Then to build and install
+locally:
+
+.. code-block:: bash
+    # Where $CONDA_FORGE_REPOSITORY could be, for example,
```
wxtim, line 183: Is this valid .rst? I thought that you would need another line here?

docs/howto/create-a-release.rst
```diff
@@ -102,6 +102,115 @@ Upload build to PyPI
...
+To test packages locally, first you should make sure that Conda is
+configured to avoid automatically upload the package. Open you ``~/.condarc``
+and check that you have something similar to:
+
+.. code-block:: yaml
+    channels:
+      - defaults
+      - conda-forge
+    ssl_verify: true
+    anaconda_upload: false
```
wxtim, line 176:
```suggestion
.. code-block:: yaml

   channels:
     - defaults
     - conda-forge
   ssl_verify: true
   anaconda_upload: false
```
Sorry - rst is terribly picky about whitespace but I think that this should fix it - check at
https://github.com/kinow/cylc-admin/blob/conda-forge-instructions/docs/howto/create-a-release.rst

docs/howto/create-a-release.rst
```diff
@@ -102,6 +102,115 @@ Upload build to PyPI
...
+    # This will take some minutes and print useful information.
+    conda build recipe/
+
+    # The following command will install the locally created package. Before
+    # installing it will ask you to confirm. Scroll up and search the
+    # package name. The right-side column must show a location like
+    # .../anaconda3/conda-bld/linux-64::cylc-uiserver-0.1-py37_1.
+    # This confirms you are installing the local build. Here $PACKAGE_NAME
+    # could be something like cylc-uiserver.
+    conda install $PACKAGE_NAME
```
wxtim, line 199:
```suggestion
.. code-block:: bash

   # Where $CONDA_FORGE_REPOSITORY could be, for example,
   # cylc-uiserver-feedstock.
   cd $CONDA_FORGE_REPOSITORY
   # Your package should not be listed!
   conda list
   # This will take some minutes and print useful information.
   conda build recipe/
   # The following command will install the locally created package. Before
   # installing it will ask you to confirm. Scroll up and search the
   # package name. The right-side column must show a location like
   # .../anaconda3/conda-bld/linux-64::cylc-uiserver-0.1-py37_1.
   # This confirms you are installing the local build. Here $PACKAGE_NAME
   # could be something like cylc-uiserver.
   conda install $PACKAGE_NAME
```

docs/howto/create-a-release.rst
```diff
@@ -102,6 +102,115 @@ Upload build to PyPI
...
+other conda packages as dependencies!.
+- Check if you need to update the version in the Jinja2 variable right
+at the top of the file.
+- If you updated the version, set the build number back to 0.
+- If you did not update the version, increase the build number.
+- If the conda package contains a source URL, check that it is evaluated
+correctly after the version is replaced, and update the ``sha256`` value.
+- Check requirements (we may have changed setup.py, or package.json, etc.
+ensure we are up to date).
+- Ensure the test commands are still valid
```
wxtim, line 147: What marks them out as valid?
What should they do?

kinow: Added the following

>- Ensure the test commands are still valid (i.e. they can be executed successfully)

As the commands are YAML strings that may represent Python commands, Python imports, shell commands, etc. So a developer releasing Cylc UI Server should see a test command such as 

```
test:
  imports:
    - cylc.uiserver
```

and could either test it with `python -c 'import cylc.uiserver'` after the package was installed locally, or simply confirm that we have not moved/renamed packages (I would do the latter).

docs/howto/create-a-release.rst
kinow: >how do I carry out this check?

I do that mentally or with a notepad. The `meta.yaml` is parsed first by a Jinja processor. If you look at the top of the file, there are variables declared (we are using the template from Conda Forge, so 99% of other recipes in Conda Forge should look similar).

The way I do it, is opening a notepad, pasting that URL you linked above, then manually replacing the variables (a manual `eval` ? :smiley: ). Then paste it into the browser, and confirm I don't get a 404. I added this step because it happened once to me that I had a wrong value, and PYPI returned 404. Finding that later wasn't really easy.

I've changed the text to 

>- If the conda package contains a source URL, check that it is evaluated
correctly using the Jinja variables, and update the ``sha256`` value.

WDYT?

>How do I generate the hyperlink and how do I generate the new checksum?

To be honest, I never do that for the Python dependencies. I simply go to https://pypi.org, search for something like `metomi-isodatetime`, then click on "Download Files". At the right side, there's a button to show the hashes, which includes the value used here.

Alternatively, I think the release-manager dev could prefer to download the file and run something like `sha256sum $file`.

I've added the following now to the end of the text

>- If the conda package contains a source URL, check that it is evaluated
correctly using the Jinja variables, and update the ``sha256`` value
(you can get this value from PYPI, or use some tool like ``sha256sum``).

docs/howto/create-a-release.rst
kinow: >Suggest that you add indication of current structure of dependencies:

Maybe that could be a follow-up change? And I think this should be for both PYPI and Conda, as I think they would share the order/dependency relationship? What do you think?

>Am I correct in thinking that this is github.com/conda-forge/${REPO}/blob/master/recipe/meta.yaml for each of the repos shown below?

Yup! The link for each repository we have is listed below this text. Should I remove that, or modify it perhaps? Below the list, it points that the developer releasing it should look for `recipe/meta.yaml` (assuming s/he cloned the repo).

kinow:
Thanks for the feedback @wxtim ! Updated with some commits, but replied a couple questions with other questions :)

docs/howto/create-a-release.rst
wxtim: That would have been my guess without futher clarification, so I'm happy w' that.

docs/howto/create-a-release.rst
wxtim: Links are fine - I just need to read more carefully.
Perhaps:
```rst
- https://github.com/conda-forge/metomi-isodatetime-feedstock (`mainly the recipe/meta.yaml file <https://github.com/conda-forge/${REPO}/blob/master/recipe/meta.yaml>`_)
```

docs/howto/create-a-release.rst
wxtim: > If the conda package contains a source URL, check that it is evaluated
correctly using the Jinja variables, and update the `sha256` value.

suggestion: 
>  If the conda package contains a source URL, check that it is evaluated
correctly by manually replacing the Jinja variables, and trying this out in a browser. 
Then update the `sha256` value. (you can get this value from PYPI, or use some 
tool like sha256sum).

kinow: Suggestion applied! :tada: thanks Tim!

docs/howto/create-a-release.rst
kinow: I'll leave it as-is as I couldn't find an easy way to add the links in every list item :smile: but hopefully the paragraph below with the location of `meta.yaml` should help for now.

kinow: BTW, one difference with conda is that it has two extra packages not present in PYPI, `cylc-ui` (JS) and `cylc` (metapackage with links to each dependency in the Cylc 8 release). If you share the SVG later or add it in a PR I can try to add these two too.

PR #89, opened 2020-04-28
SOD Proposal doc: addressed Dave's feedback.

hjoliver:
@dpmatthews - putting this up as a PR so you can see the specific changes more easily (this is in response to our email discussion of a week or so ago).  I think I've addressed everything we covered.
That makes this document pretty comprehensive. I'd like to go back and make it is as concise as possible but probably not best use of my time at this point (I'll refine it for documentation purposes later on, once SoD is official :grin: ).

### DIFF
--- a/docs/proposal-spawn-on-d.md
+++ b/docs/proposal-spawn-on-d.md
@@ -18,13 +18,13 @@
   - [Spawning on task completion](#spawning-on-task-completion)
   - [Failed tasks](#failed-tasks)
   - [Stall and shutdown](#stall-and-shutdown)
-  - [Suicide triggers not needed](#suicide-triggers-not-needed)
+  - [Suicide triggers mostly not needed](#suicide-triggers-mostly-not-needed)
   - [Intervention and reflow](#intervention-and-reflow)
      - [Reflow in SoS](#reflow-in-sos)
      - [Reflow in SoD](#reflow-in-sod)
   - [Submit number](#submit-number)
   - [CLI implications](#cli-implications)
-  - [Succeeded tasks](#succeeded-tasks)
+  - [Succeeded vs failed tasks](#succeeded-vs-failed-tasks)
 - [Notes](#notes)
 - [Future enhancments](#future-enhancements)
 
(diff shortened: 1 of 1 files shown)


docs/proposal-spawn-on-d.md
```diff
@@ -56,46 +58,51 @@ and usage, and will make Cylc 9 changes easier to implement in the future.
...
-parents have finished *or* the pool has moved beyond the point it can affect them
-1. [Shut down](#scheduler-shutdown) if stalled with no failed tasks present
+1. [Remove waiting tasks](#housekeeping-waiting-and-finished-tasks) if all of
+      their parents have finished
+1. [Remove finished tasks](#housekeeping-waiting-and-finished-tasks) if all of
+      their parents have finished *or* if the pool has moved beyond the point
+      it can affect them
+1. [Shut down](#stall-and-shutdown) if stalled with no failed tasks present
 
 ## Discussion of details
```
dpmatthews, line 107: `(and maybe [unhandled succeeded  tasks](#succeeded-tasks)?)`
This bit (in the description of the task pool below) is now a broken link (and we've now said no to this anyway).

docs/proposal-spawn-on-d.md
```diff
@@ -631,74 +674,84 @@ In SoD we need to glob on:
...
+- the parent_finished status populated so far
 
-Also need to auto-spawn tasks with no prerequisites again, like at cold start.
+### Reload and Restart after graph changes
 
-Similarly for reload, if parentless task defs were added.
+This gets much better under SoD but is still a little difficult. If new task
+definitions are added, they will automatically be spawned on demand so long as
+existing tasks are their parents in the graph. But if they have no parents we
+will still have to insert the first instance at the desired cycle point.
```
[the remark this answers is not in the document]
hjoliver: I didn't put that in because it could still be difficult to choose the actual cycle point, unless perhaps you mean the "first valid cycle point at or beyond the current runahead limit point"? (Note that all tasks are initially spawned into the runahead pool, even if their cycle point is not beyond the limit). For instance, the oldest or newest point in the current pool might seem sensible, but the suite could be very widely spread across cycle points.  However, I'll add a note that we should consider something of that nature...

docs/proposal-spawn-on-d.md
hjoliver: (Removed that broken bit).

hjoliver:
Feedback addressed; merging...

PR #90, opened 2020-04-29
Update config changes proposal based on CylcCon discussion

dpmatthews:
See https://cylc.github.io/cylc-admin/feb2020-workshop-notes#wednesday

### DIFF
--- a/docs/proposal-config-changes.md
+++ b/docs/proposal-config-changes.md
@@ -1,29 +1,31 @@
 # Proposal - changes to global and workflow configs
 
 These notes mainly address [cylc-flow #3422](https://github.com/cylc/cylc-flow/issues/3422).
+They have been updated to reflect the discussion at
+[CylcCon 2020](https://cylc.github.io/cylc-admin/feb2020-workshop-notes#wednesday).
 
 ## General points
 
 ### Terminology
    * We should remove all references to `suite` in the settings (partly because
      we're trying stop referring to "suites" and partly because some affected
-     settings are actually referring to the workflow server which is unclear).
-     * Note: "suite" is referred to through-out the code, tests and
-       documentation.
-   * **NOTE** "server" changed to "scheduler" below in light of the recent
-     component name decision - *we still need to check each instance carefully*
+     settings are actually referring to the workflow scheduler which is unclear).
+   * We will refer to the process which controls the workflow as the "scheduler"
+    (previously referred to as either "daemon", "server" or "service").
+   * **NOTE**: The old terminology is still used through-out the code, tests and
+     documentation - these will need to be updated (not urgent).
 
 ### Deprecation
    * The global config will be changing drastically for the platforms support.
      Furthermore, it doesn't affect most end users.
-     * Proposal: Don't support any deprecated or obsolete settings in the global
+     * Decision: Don't support any deprecated or obsolete settings in the global
        config.
    * For the workflow config the code to support deprecated and obsolete settings
      isn't complicated so there is no big overhead for us. On the other hand we
      don't want users using old syntax for too long. Some settings only became
      deprecated or obsolete very recently - it seems harsh to remove support for
      these.
-     * Proposal: Remove everything from 7.6.0 (Feb 2018) and before. That will
+     * Decision: Remove everything from 7.6.0 (Feb 2018) and before. That will
        just leaves things from 7.8.x.
 
 ### Config file naming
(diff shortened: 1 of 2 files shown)


docs/proposal-config-changes.md
```diff
@@ -392,25 +396,49 @@ Matches global config change to `[task events]mail retry delays`.
...
+
+### `[runtime][]extra log files` -> obsolete
+
+Instead provide access to all files found in the job log dir in the new UI. See
+[discussion](https://github.com/cylc/cylc-flow/pull/3348#discussion_r334219315).
+
+### Empty string defaults
+
+Where default values are currently defined as empty strings, these will be
+replaced with `None`. See
```
dpmatthews, line 433: Actually I'm not sure about this. Some config settings don't have a default defined whereas others have an empty string ('') - what is the difference?

docs/proposal-config-changes.md
```diff
@@ -392,25 +396,49 @@ Matches global config change to `[task events]mail retry delays`.
...
+`cycle point num expanded year digits` and `cycle point time zone` (discussed at
+CylcCon).
+
+### `[cylc][simulation]disable suite event handlers` -> ???
+
+Agreed at CylcCon to move this and make it positive.
+Notes suggest `[scheduling]use suite event handlers`.
+However current setting only applies to simulation and dummy mode (and defaults to true) -
+we don't need this setting for real workflows do we (and we want these handles disabled by default in simulation and dummy mode)?.
+How about `[scheduler][events]enable simulation handlers`?
```
dpmatthews, line 423: Not sure about this bit

docs/proposal-config-changes.md
```diff
@@ -225,10 +245,10 @@ SPEC = {
     'monitor': {
         'sort order': [VDR.V_STRING, 'definition', 'alphanumeric'],
     },
-    'job platforms': {
+    'platforms': {
         '__MANY__': {
-            'batch system': [VDR.V_STRING, 'background'],
-            'batch submit command template': [VDR.V_STRING],
+            'job runner': [VDR.V_STRING, 'background'],
```
oliver-sanders, line 250: Quick question, do we want to refer to the default "job runner" as "background" or None?

docs/proposal-config-changes.md
oliver-sanders: At the moment if you don't specify a default you get `None` though there is some logic to provide `[]` if that is more appropriate.

I have tweaked this a little in https://github.com/cylc/cylc-flow/pull/3559, now the configuration can tell the difference between a setting with no default and one with a default, though unset defaults still get expanded to `None` or `[]`.

oliver-sanders: I'm not in any hurry to replace `''` with `None` though it would be nicer if all unset values defaulted to `None`.

datamel: Happy to be corrected but from my understanding None is nonetype - not the same as an empty string. If it is going to be used for comaprison, 'none' means not yet set (or no value passed to the optional parameter), whereas "" might be a valid option.
From reading around None seems to be used by default more frequently. Although if you want to use string methods on it (regarless of whether it is there or not), then "" is better as if you use a sting method on None it will throw exceptions.
I think I remember reading somewhere that None can be good as if it isn't there isn't any memory assigned to it to it, whereas "" would - can't confirm this though - again happy to be corrected!

oliver-sanders: That's right it's just a matter of how we use the settings.

Using empty strings as default values makes things a little easier in some ways:

```python
cmd = f'{ cfg["command-prefix"] } mail { cfg["mail opts"] } -s foo'
```

Whereas when using `None` defaults:

```python
cmd = f'{ cfg["command-prefix"] or "" } mail { cfg["mail opts"] or "" } -s foo'
```

docs/proposal-config-changes.md
```diff
@@ -195,28 +219,24 @@ SPEC = {
...
             'method': [VDR.V_STRING, 'name', 'address', 'hardwired'],
             'target': [VDR.V_STRING, 'google.com'],
             'host': [VDR.V_STRING],
         },
+        'run hosts': {
+            'available': [VDR.V_SPACELESS_STRING_LIST],
+            'condemned': [VDR.V_ABSOLUTE_HOST_LIST],
+            'ports': [VDR.V_INTEGER_LIST, list(range(43001, 43101))],
+            'ranking': [VDR.V_STRING],
+        },
```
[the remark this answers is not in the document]
dpmatthews: We need to work out the details of how the ranking will work in a platform group so I suggest we deal with this in a separate PR / issue.

docs/proposal-config-changes.md
hjoliver: #3508 resulted from getting this sort of thing wrong (although that particular example will be irrelevant in the "Platforms" era.   Sometimes you need to know if an item has been explicitly been set to "nothing" (the exact value of which depends on its type) or not set at all.

hjoliver: Sounds like this convo is resolved (unresolve if not!)

docs/proposal-config-changes.md
hjoliver: Would `None` mean you can't run jobs (i.e. users have to set an explicit value or their jobs won't run), or would it mean "background"?   I can see an argument both ways, but having to be explicit would making writing dummy test suites a little bit more difficult.

docs/proposal-config-changes.md
```diff
@@ -88,16 +78,11 @@ name. See
 See
 [discussion](https://github.com/cylc/cylc-flow/pull/3348#discussion_r333229597).
 
-### `[cylc]task event mail interval` -> `[scheduler][mail]task event interval`
+### `[cylc]task event mail interval` -> `[scheduler][mail]task event batch interval`
```
hjoliver, line 81: This conforms with my meeting notes, but shouldn't it be `task event mail batch interval`, for complete clarity?  (We don't do any other kind of event batching do we?).

docs/proposal-config-changes.md
hjoliver: Shall we just remove it and automatically disable event handlers in dummy and sim modes?  It's easy to test event handlers, if you need to, in a real test suite written for that purpose.  I can't really think of a good use case for this when running a real suite in dummy or sim mode.

docs/proposal-config-changes.md
oliver-sanders: I was thinking that `None` would mean `background`.

hjoliver: `None` could also be better on grounds that "background" is not strictly correct (in the shell "background process" sense - done with `&` and `wait`able).   It more (as you say) that we run the job directly without submitting to an external job management system.

docs/proposal-config-changes.md
dpmatthews: It's already in the `[mail]` section- is that not enough?

docs/proposal-config-changes.md
dpmatthews: Agreed - I've updated the proposal to make this setting obsolete

docs/proposal-config-changes.md
dpmatthews: Personally I'd prefer to stick with "background", at least for the moment. In the future it won't really matter since users will only need to worry about platforms and you should never need to specify "job runner = background" in the global config.

hjoliver: OK, let's leave it for now.

docs/proposal-config-changes.md
hjoliver: Ah, sorry, missed that. Duh!

PR #93, opened 2020-07-17
Fix markdown link in Cylc 8 release and typos

kinow:
Remembered that when reading the Conda notes this morning, didn't find the `--use-local` in the docs.

Reading the release notes in our site, noted that one markdown link (what's new) was broken.

And the IDE found some typos, so just applied its suggestions.

:+1: nothing critical here, doesn't need to be reviewed or merged before our release.

### DIFF
--- a/docs/howto/create-a-release.md
+++ b/docs/howto/create-a-release.md
@@ -194,7 +194,7 @@ locally:
    # .../anaconda3/conda-bld/linux-64::cylc-uiserver-0.1-py37_1.
    # This confirms you are installing the local build. Here $PACKAGE_NAME
    # could be something like cylc-uiserver.
-   conda install $PACKAGE_NAME
+   conda install --use-local $PACKAGE_NAME
 ```
 
 At this point you should be good to go. Test the package with commands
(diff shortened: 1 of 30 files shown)


docs/howto/create-a-release.md
```diff
@@ -194,7 +194,7 @@ locally:
    # .../anaconda3/conda-bld/linux-64::cylc-uiserver-0.1-py37_1.
    # This confirms you are installing the local build. Here $PACKAGE_NAME
    # could be something like cylc-uiserver.
-   conda install $PACKAGE_NAME
+   conda install --use-local $PACKAGE_NAME
```
kinow, line 197: This is useful as it will try to use the locally built version of the package, like `cylc-uiserver`. If one forgets this flag (as I did this morning :grimacing:) you still have a chance to confirm whether you want to proceed or not. But no harm in documenting it here too.

docs/meetings/vc-4-apr-2019-summary.md
```diff
@@ -24,7 +24,7 @@ workflow.
 
 **ACTION: Martin to put WIP code on GitHub and make it available to others** - DONE
 
-Ascyncio in WS: not done yet, but not on the critical path (can be done later).
+Asyncio in WS: not done yet, but not on the critical path (can be done later).
```
kinow, line 27: These meetings notes are internal only, but thought it was also harmless to update since the IDE did all the hard work to locate/change the text.

PR #94, opened 2020-07-17
add instruction for testing the docs

oliver-sanders:
Because cylc-doc autodocuments other projects it is necessary to test that the docs build correctly before releasing any of these documented projects. Otherwise we will have to re-release in order to fix little things like broken urls.

These breakages should get picked up by our nightly testing (when we get that off the ground), however, this is a quick simple check which should be worthwhile.

### DIFF
--- a/docs/howto/create-a-release.md
+++ b/docs/howto/create-a-release.md
@@ -53,6 +53,19 @@
    rm -r dist
 ```
 
+## Test the docs
+
+For any projects which are auto-documented by cylc-doc, currently:
+
+* cylc-flow
+
+Ensure the docs build against master by manually triggering the test workflow
+in cylc-doc.
+
+This will catch syntax errors, broken urls etc which need to be fixed
+prior to releasing the project.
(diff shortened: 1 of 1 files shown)


docs/howto/create-a-release.md
```diff
@@ -53,6 +53,19 @@
...
 ```
 
+## Test the docs
+
+For any projects which are auto-documented by cylc-doc, currently:
+
+* cylc-flow
+
+Ensure the docs build against master by manually triggering the test workflow
+in cylc-doc.
```
wxtim, line 63: Can you put a tad more detail in for the unfamiliar?
Where and how do I do this?

oliver-sanders: I'm reluctant to document how to use GH actions, partly because GH does a better job of documenting this but also as the instructions will continue to change, only a few weeks ago the actions tab was in a completely different place and workflow_dispatch didn't even exist.

PR #95, opened 2020-07-17
platforms proposal: make installation configurable at the suite level

dpmatthews:
Also fix a mistake in the config changes proposal.

### DIFF
--- a/docs/proposal-platforms.md
+++ b/docs/proposal-platforms.md
@@ -286,32 +286,44 @@ suite-run functionality:
        * Hopefully it should be fine to install the same files to all install
          targets. If there are cases where this is an issue we can recommend the
          use of install tasks as a workaround.
+       * `rose suite-run` currently installs everything other than a set list of
+         exclusions. However, it has been known for suites to write to top level
+         directories (possibly not deliberate) so this is not a good default
+         behaviour (especially with the `--delete` rsync option).
        * Proposal:
          * Install the same items to all install targets.
          * Allow the installation items to be configured at the suite level via
-           ones of the following ways:
-           * A new setting `[scheduler]install` with a default value of
-             `app bin etc lib` (the items in `.service` are always installed).
-             Note that if we want to be able to support copying top level files
-             as well as directories then we would need to do something like
-             `app/ bin/ etc/ lib/ extra-dir-to-copy/ extra-file-to-copy`.
-           * The alternative is to support the use of an extra file
-             `.rsync-filter` where you can specify rsync filter rules. All we
-             would need to do in Cylc is use the rsync option
-             `--filter=': .rsync-filter'`. To configure an additional directory
+           either of the following methods:
+           * A new setting `[scheduler]install` (with a default value of None)
+             where you can specify extra top level directories and files to be
+             installed, e.g. `extra-dir-to-copy/ extra-file-to-copy`
+             (directories must have a trailing slash). Patterns are allowed (as
+             defined by rsync), e.g `*/ *` would mimic the current
+             `rose suite-run` behaviour.
+           * Create an extra file in the suite directory named
+             `.rsync-filter` where you can specify rsync filter rules. For
+             example, to configure an additional directory
              "data" to be copied you would simply create the file
              `.rsync-filter` in the top level directory and add the line
              `+ /data/***`. We can document a simple example but refer users who
-             want to do anything complicated to the rsync man page. Advantages:
+             want to do anything complicated to the
+             [rsync man page](https://linux.die.net/man/1/rsync). Advantages:
              a) gives users access to the full power of rsync filters;
              b) this file can be added to cylc 7 suites without breaking them.
-     * The alternative is to use exclude rather than include and make sure we
-       exclude any file / directories which are generated at run time or target
-       specific (`log/ share/ work/` + various files / directories in
-       `.service/`).
-       * There are some suites around which write to top level directories (not
-         sure if this is deliberate) - exclude would be very bad in these cases.
-         Therefore, include is the safer option.
+         * The rsync filter options will be applied in the following order (note
+           that the first matching pattern is acted on):
+           * Include files required by cylc:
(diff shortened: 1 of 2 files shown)


docs/proposal-platforms.md
```diff
@@ -277,10 +278,33 @@ suite-run functionality:
...
+         use of install tasks as a workaround.
+       * Proposal:
+         * Install the same items to all install targets.
+         * Allow the installation items to be configured at the suite level via
+           ones of the following ways:
+           * A new setting `[scheduler]install` with a default value of
+             `app bin etc lib` (the items in `.service` are always installed).
+             Note that if we want to be able to support copying top level files
+             as well as directories then we would need to do something like
+             `app/ bin/ etc/ lib/ extra-dir-to-copy/ extra-file-to-copy`.
```
oliver-sanders, line 297: Note that `[scheduler]install` should not accept patterns or anything other than top-level files or directories.

hjoliver: Why is that exactly?

dpmatthews:
@hjoliver - @datamel is working on this now so we need to at least decide what to implement initially.
I'd be happy to either support both or possibly just support the rsync filter initially (which is trivial to implement) and then add the config item later once we've had a chance to try out the functionality (if we decide we need it).

docs/proposal-platforms.md
oliver-sanders: 1) Force adoption of standard permitted directories (`bin`, `lib`, `app`).
2) Prevent overly generic globs (e.g. `*`).
3) Multi-level includes would require extra logic in forming the `rsync` command.

datamel:
Can I confirm that the intention is, when the `[scheduler]install` setting is present in a suite.rc, the default values of `app bin etc lib` will all be overridden? (and that the .service is essentially hard-coded / non-configurable).

I ask because:

> Note that if we want to be able to support copying top level files as well as directories then we would need to do something like app/ bin/ etc/ lib/ extra-dir-to-copy/ extra-file-to-copy.

suggests that we would need to allow additional files and directories to be appended i.e. would we then expect the user to 'spell out' those defaults themselves to be able to add additional sources? i.e. in what scenario will these default 'abel' :) folders be overridden; the moment a new source is added in `[scheduler]install`?

I agree with @oliver-sanders that we can only really offer top-level sources to be added in the suite.rc. Any depeer would presume some rsync knowledge, which wouldn't be intuitive in a non-rsync config file (and to all users) e.g. to add a file deep in you also need to add the directories it is under separately, also.

I like the idea of supporting both. They play well together although the order on command line is important. We would typically keep the `exclude="*"` at the end but putting the filter before includes would mean filters take prescedence. Again, we need to clarify how this sits then, with the 'abel' folders, which will be coming in via `--include`.

We just need to be clear about what 'trumps' what between defaults, suite.rc and the filters. Here in the proposal and then in the documentation.

Useful to note that `-F` can be used in place of `--filter=': .rsync-filter'` and `-FF` will also exclude the .rsync-filter itself from being copied.

dpmatthews:
> Can I confirm that the intention is, when the `[scheduler]install` setting is present in a suite.rc, the default values of `app bin etc lib` will all be overridden? (and that the .service is essentially hard-coded / non-configurable).

I was suggesting that just in case anyone wanted to remove some of the defaults but that's not needed if we support the rsync filter as well.

> I like the idea of supporting both. They play well together although the order on command line is important. We would typically keep the `exclude="*"` at the end but putting the filter before includes would mean filters take prescedence. Again, we need to clarify how this sits then, with the 'abel' folders, which will be coming in via `--include`.

I think the filter should take precedence (so that the users can override the defaults if they want to). I'll update the proposal.

docs/proposal-platforms.md
dpmatthews: I can't see any reason not to allow patterns so long as we explicitly exclude anything which should never be copied. This makes it easy for users to mimic the current `rose suite-run behaviour`. See updated proposal.

dpmatthews: Resolving this - please comment on the updated proposal

docs/proposal-platforms.md
```diff
@@ -286,32 +286,44 @@ suite-run functionality:
...
-         sure if this is deliberate) - exclude would be very bad in these cases.
-         Therefore, include is the safer option.
+         * The rsync filter options will be applied in the following order (note
+           that the first matching pattern is acted on):
+           * Include files required by cylc:
+             `--include='/.service/' --include='/.service/server.key'`
+           * Exclude directories which should never be copied:
+             `--exclude='.service/***' --exclude='log' --exclude='share' --exclude='work'`
+           * Apply any rules define in an `.rsync-filter` file:
+             `--filter=': .rsync-filter'`
```
datamel, line 320: Would you like users to specify in their suite design that they are using a filter file, or do you just want Cylc to check and use the filter file if one happens to exist?

dpmatthews: I'm suggesting that cylc always uses this rsync option for the installation and then rsync will use the file if it exists.

docs/proposal-platforms.md
```diff
@@ -277,17 +278,52 @@ suite-run functionality:
...
-         sure if this is deliberate) - exclude would be very bad in these cases.
-         Therefore, include is the safer option.
+       * If we allow global configuration this is likely to lead to portability
+         issues where a suite only works if you modify your global config. If
+         additional items need to be installed it's better to force these to be
+         configured in the suite. Note that this will require changes to some
+         Cylc 7 suites in order for them to work at Cylc 8.
+       * Hopefully it should be fine to install the same files to all install
+         targets. If there are cases where this is an issue we can recommend the
+         use of install tasks as a workaround.
```
hjoliver, line 288: Not entirely sure about this, but its fine as a first cut.

dpmatthews:
All feedback addressed so merging this. We can fine tune if necessary as part of the implementation.

PR #97, opened 2020-07-23
Include information for future releases about troubleshooting conda releases

kinow:
Hopefully it will be helpful for future release-managers 👍

### DIFF
--- a/docs/howto/create-a-release.md
+++ b/docs/howto/create-a-release.md
@@ -114,6 +114,8 @@ A may have multiple dependants, but no dependency to other modules).
 For example, Cylc UI Server depends on Cylc Flow. So unless you are
 releasing only Cylc UI Server, you should release Cylc Flow first.
 
+### Prepare the Conda Releases
+
 On GitHub, navigate to the project repository on GitHub, e.g.
 
 - https://github.com/conda-forge/metomi-isodatetime-feedstock
@@ -162,6 +164,8 @@ to show in Conda Forge. For example, if Azure or GitHub infrastructure
 have issues (as I experienced during the first releases) it may take days.
 So keep that in mind before announcing the Conda packages.
 
+### Testing the Packages
+
 To test packages locally, first you should make sure that Conda is
 configured to avoid automatically upload the package. Open your `~/.condarc`
 and check that you have something similar to:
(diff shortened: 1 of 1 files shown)


docs/howto/create-a-release.md
```diff
@@ -202,14 +206,62 @@ such as `which $PACKAGE_NAME`, or `$PACKAGE_NAME --version`, etc.
...
+#### Marking a release as invalid or broken
+
 To undo a release, you will have to liaise with the maintainers via Gitter
 or Github. Or, alternatively, bump up the build number +1, and release it
 again.
 
+#### `conda-install` is picking the wrong version or build of a package
+
+If `conda install cylc` returns the incorrect build number, it could be
+an issue that is very difficult to debug (a posteriori).
```
hjoliver, line 226:
```suggestion
an issue that is very difficult to debug (a posteriori). In fact it may not even be a bug:

Question to conda Gitter channel: (@hjoliver Jul 21 17:23):
> I have a conda metapackage just published to conda-forge, which (on conda install) does not install the latest build numbers of several of our component packages. The conda docs cite bug fixes as a primary use case for incrementing build numbers, in which case I would have thought the environment solver should never select older builds? If I install the component packages directly, I do get the latest builds, but not via the metapackage. Anyone know what's going on here?

Replies:
> @wolfv:  
@hjoliver this article describes how the conda solver works: https://www.anaconda.com/blog/understanding-and-improving-condas-performance. It specifically says that build numbers are only maximized for explicit specs

> @jjhelmus: 
@hjoliver changes to the build number can also occur when a package is built against a different version of a library which is incompatible with earlier version. For example, one build of python 3.7.7 might be compiled against libffi 3.2 while the next build number against libffi 3.3. Depending on what other packages are installed in the environment it may not be possible to install the package with the largest built number. In the example if cffi is only available as a package built against libffi 3.2 then the older build of python 3.7.7 would be required.

Interpretation: build number is treated much like version number, if not pinned down exactly conda can choose an older build if that works best for solving all the dependencies in the system. Conda would presumably(?) choose the newest build if it does not affect dependencies; we may just need to check what happens if dependencies get changed by a new build. 
```

kinow: Thanks!!! :rocket:

PR #99, opened 2020-07-27
howto: run the tests

wxtim:
A document knocked together to give new starters a leg up debugging tests. It was originally written for a system which could interpret RST. :( 
I don't have time to convert it to GH markdown, but I think that it still has value.


docs/howto/testing.rst
```diff
@@ -0,0 +1,181 @@
...
+
+================
+How to run tests
+================
+
+
+All tests are in the folder ``cylc-flow/tests``. Instructions
+here, unless stated otherwise assume that you are working from
+your cylc-flow repo dir.
+
```
datamel, line 13: Worth putting a comment in here about flow-tests.rc file?

hjoliver: Agreed!

wxtim: Do we have an example of flow tests.rc?

PR #100, opened 2020-07-28
SoD proposal/doc update.

hjoliver:
Two new commits.
- The first commit documents exactly how suicide triggers work under SoD (after an email from @dpmatthews)
- The second commit documents the new handling of absolute triggers (after SoD review comments from @dwsutherland)

Both add minor items to the TODO list, to be converted to issues after the SoD PR is merged.

### DIFF
--- a/docs/proposal-spawn-on-d.md
+++ b/docs/proposal-spawn-on-d.md
@@ -216,8 +216,8 @@ spawn an infinite number of children (`bar.1,2,3,...`), which obviously is not
 feasible.
 
 (Note without the first line above `start` is not defined on any sequence, so
-with SoS the scheduler will stall with `foo` unsatisfied, but with SoD it will
-shut down immediately with nothing to do because and `foo` never gets spawned.)
+with SoS the scheduler will stall with `bar` unsatisfied, but with SoD it will
+shut down immediately with nothing to do because and `bar` never gets spawned.)
 
 Worse, `bar` could also have other non-absolute triggers, and it could be
 those that spawn initial instances of `bar` before `start` is finished:
--- a/docs/proposal-spawn-on-d.md
+++ b/docs/proposal-spawn-on-d.md
@@ -205,22 +205,51 @@ trigger events too, but for now a waiting task with an xtrigger is fine).
 
 ### Absolute dependence
 
-Here `foo.1,2,3,...` should trigger off of `start.2`:
+At first glance this is not a great fit for the spawn-on-demand model: 
 ```
 ...
       R1/2 = "start"
-      P1 = "start[2] => foo"
+      P1 = "start[2] => bar"
 ```
+Above, with no final cycle point a single event (`start.2 succeeded`) should
+spawn an infinite number of children (`bar.1,2,3,...`), which obviously is not
+feasible.
+
 (Note without the first line above `start` is not defined on any sequence, so
 with SoS the scheduler will stall with `foo` unsatisfied, but with SoD it will
 shut down immediately with nothing to do because and `foo` never gets spawned.)
 
-So: when an absolute parent finishes *spawn its children at the first cycle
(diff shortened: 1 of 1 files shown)


docs/proposal-spawn-on-d.md
```diff
@@ -205,22 +205,51 @@ trigger events too, but for now a waiting task with an xtrigger is fine).
...
-      P1 = "start[2] => foo"
+      P1 = "start[2] => bar"
 ```
+Above, with no final cycle point a single event (`start.2 succeeded`) should
+spawn an infinite number of children (`bar.1,2,3,...`), which obviously is not
+feasible.
+
 (Note without the first line above `start` is not defined on any sequence, so
 with SoS the scheduler will stall with `foo` unsatisfied, but with SoD it will
 shut down immediately with nothing to do because and `foo` never gets spawned.)
```
wxtim, line 220:
```suggestion
with SoS the scheduler will stall with `bar` unsatisfied, but with SoD it will
shut down immediately with nothing to do because and `bar` never gets spawned.)
```

hjoliver: Good spotting @wxtim . I had changed `foo` to `bar` in the example, to be more like the new example added below it :grimacing:

dpmatthews: "because and" => "because"
`foo` should be `bar` (and 2 lines above)

oliver-sanders:
Seems legit, absolute dependencies should be fairly few (usually just one) so the overheads of this caching are pretty minimal.

docs/proposal-spawn-on-d.md
```diff
@@ -205,22 +205,51 @@ trigger events too, but for now a waiting task with an xtrigger is fine).
...
+    children already in the pool due to spawning by others (e.g. `foo` above).
+2. Whenever a task with an absolute trigger (`bar` above) is spawned, check to
+see if the trigger is already satisfied. If it is, satisfy it. If it isn't then
+it will be satisfied later when the absolute parent finishes and updates all
+child instances present in the pool as described just above. 
+3. Whenever a task with an absolute trigger is released from the runahead
+pool spawn its next instance (strictly speaking this is only necessary for
+tasks with no non-absolute triggers to spawn them on demand).
+
+#### Retriggering
```
dpmatthews, line 252: "retrigger the absolute parent without reflow, then retrigger with reflow the first child that you want to continue from"
That will only run that single child won't it?

docs/proposal-spawn-on-d.md
dpmatthews: Updated version still says "bacause and" ?

docs/proposal-spawn-on-d.md
hjoliver: - "retrigger the absolute parent without reflow" - this will only trigger the absolute parent task
- "retrigger with reflow the first child that you want to continue from" - this will trigger "the first child that you want to continue from" and that will (on release from the runahead pool) spawn the next child, and so on - because that's how we spawn tasks with absolute parents. (Note I've said "spawn" rather than "run" for subsequent child tasks).

hjoliver: I'll make the text clearer...

docs/proposal-spawn-on-d.md
hjoliver: Fixed.

hjoliver:
Comments addressed.

hjoliver:
> Seems legit, absolute dependencies should be fairly few (usually just one) so the overheads of this caching are pretty minimal.

Caching the trigger results is the SoD version of keeping absolute parents in the task pool forever in SoS.

hjoliver:
Having addressed the comments and gotten one approval I'm going to merge this so that the document is up to date with the just-merged implementation PR. (We can always tweak the docs again later).

PR #101, opened 2020-07-29
Add cylc wrapper warning to a2 release notes.

hjoliver:
Tell users how to test Cylc 8 without a central wrapper script (local background jobs) and under what circumstances they may need to disable their Cylc 7 wrapper script to do so.


hjoliver:
@dpmatthews and @oliver-sanders - in light of recent discussions please check you agree with the text here.

oliver-sanders:
Ok, lets try and document a Cylc7/8 compatible wrapper for Conda installations in the 8.0a3 release.

PR #102, opened 2020-07-29
interfaces and protocols

oliver-sanders:
Add a diagram showing the history and progression of communication interfaces and protocols along with a breakdown of the bits still to be done.


oliver-sanders:
> Generally good.. Debatable items in the remaining work, but mostly agree 👍

Thought so, best flush these out on Element.

PR #105, opened 2020-08-07
how to debug pytest despite distribution

wxtim:
How to debug python tests.

### DIFF
--- a/docs/howto/testing.md
+++ b/docs/howto/testing.md
@@ -50,6 +50,20 @@ You may find the following idiom useful in a `flow-test.rc` file:
 ```
 
 
+Debugging Pytests
+=================
+
+You should first run tests in verbose mode using ``-vv``.
+
+If you wish to use ``breakpoint()`` to investigate failures then you will need
+to add the following to your pytest command ``pytest <path> --dist=no -n0`` to
+turn off the test distribution mode.
+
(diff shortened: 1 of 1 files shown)


oliver-sanders:
Heck, didn't know we had testing documentation here.

It think it would probs be better to copy this information into the README files in `cylc-flow/tests/` as these docs are highly likely to get out of date as they are in the wrong repo and contributors are unlikely to find them.

hjoliver:
> It think it would probs be better to copy this information into the README files in cylc-flow/tests/ as these docs are highly likely to get out of date as they are in the wrong repo and contributors are unlikely to find them.

I agree.  Do you want to do that @oliver-sanders ? (As recent author of said README files).

(BTW as far as documentation goes, cylc-admin can be used as a temporary holding tank pending decisions on final resting place).

wxtim:
> Heck, didn't know we had testing documentation here.

We didn't - this came from out internal docs (metomi-docs) and was updated for Dave. I originally wrote these for Mel distilling things you'd told me about debugging the test framework. There's nothing in these requiring them being in a private repo.

hjoliver:
> Happy to move this and replace this PR with one in cylc-flow and close this one.

@wxtim I'll merge this change here, so you have a single source to work from.

PR #108, opened 2020-08-17
platforms proposal: tweak installation method

dpmatthews:
* Remove support for rsync filter files.
  It's better to support a single way of specifying what needs to be installed via `cylc.flow`.
  rsync filter files would have allowed excludes as well as includes but we don't really need these.
  Note that cylc 7 suites which rely on the installation of directories other than `app/ bin/ etc/ lib` will need `[scheduler]install` to be defined appropriately before they will work at cylc 8.
* Allow the default set of directories to be installed to be overridden.
* The proposal remains to install the same files to all targets.
   We could consider adding support for installing on a task / platform basis later (if we decide this is required / makes sense) via a task level setting, e.g.
```ini
[runtime]
    [[task]]
        platform = myplatform
        install = ...
```

### DIFF
--- a/docs/proposal-platforms.md
+++ b/docs/proposal-platforms.md
@@ -283,46 +283,35 @@ suite-run functionality:
          additional items need to be installed it's better to force these to be
          configured in the suite. Note that this will require changes to some
          Cylc 7 suites in order for them to work at Cylc 8.
-       * Hopefully it should be fine to install the same files to all install
-         targets. If there are cases where this is an issue we can recommend the
-         use of install tasks as a workaround.
+       * We will only support installing the same files to all install targets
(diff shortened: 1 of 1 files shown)


dpmatthews:
> One question though: might we actually need custom exclude filters for the install task case - to avoid rsyncing those files? Or do you think it's OK to expect users to keep those files in a separate location?

The current proposal only allows top level directories to be be specified.
We can add support for sub-directories and exclude patterns (`(!exclude-me)`) later if we need them.
I don't recall this being asked for in all the time we've been using rose suite-run so I suggest we keep it simple for the moment.

datamel (approved):

PR #109, opened 2020-09-01
family update scheme sketch

dwsutherland:
Add the hand drawn family state update scheme/diagram included in the following PR:
https://github.com/cylc/cylc-flow/pull/3787

This documents the concept behind the event driven & recursive family state fields in illustrative form.


dwsutherland:
@hjoliver - Just added the image, is that sufficient?

hjoliver:
> @hjoliver - Just added the image, is that sufficient?

It might get lost, as a lone image with nothing linking to it.

Maybe instead of cylc-admin, a cylc-flow wiki page on documenting the data-store? That could be a dumping ground for this sort of thing, that might eventually be useful for documenting of the system's inner-workings in future?

PR #110, opened 2020-09-16
Authentication documentation

datamel:
This document explains how Cylc 8 uses CurveZMQ. This is more detail than is required for the user guide which I will update too.


dpmatthews:
How did you generate the PNG? Can you include the source in case it needs editing in future?

The PNG has the client keys named according to "PlatformName" - I think that should be InstallTarget.
Otherwise looks good to me

datamel:
> How did you generate the PNG? Can you include the source in case it needs editing in future?
> 
> The PNG has the client keys named according to "PlatformName" - I think that should be InstallTarget.
> Otherwise looks good to me

Yes, I will change that and include the source, I made it in Word (is there a preferred program to design these sorts of things?). Happy to redesign if Word is considered to be not appropriate!

hjoliver (approved):
Thanks @datamel that's great.  I'll take @dpmatthews comment as approval to merge this.

PR #111, opened 2020-10-09
Expand rose suite-run proposal to cover new cylc install command and other CLI changes

dpmatthews:


dpmatthews:
These changes already incorporate initial feedback from @oliver-sanders although I know he has further comments to make.

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,228 @@ The remaining points can happen any time after 1) is complete although note that
...
-7. On starting up a new run of the suite assign a new run name:
+7. Use a different run directory for each installation of a suite:
    ```
-   mi-aa001/
+   u-ab123/
      run1
      run2
    ```
-   * This is new functionality but Cylc 8 is a good time to introduce it.
+   * This is big change in working practise but Cylc 8 is a good time to
```
wxtim, line 86:
```suggestion
   * This is big change in working practice but Cylc 8 is a good time to
```

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,228 @@ The remaining points can happen any time after 1) is complete although note that
...
-   * Need Cylc equivalent but this should be done as part of
+   * Longer term we should build housekeeping into Cylc, see
      [cylc-flow #1159](https://github.com/cylc/cylc-flow/issues/1159).
 
-### CLI
+9. Record version control information when you install a suite, see
+   [cylc-flow #3849](https://github.com/cylc/cylc-flow/issues/3849).
+   * `rose suite-run` current supports this for Subversion only
+     (`svn info` + `svn status` + `svn diff`).
+   * Extend this to support equivalent information for git.
```
wxtim, line 100: 👍🏻

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,228 @@ The remaining points can happen any time after 1) is complete although note that
...
+
+### CLI changes
+
+`cylc install` - new command to install workflows.
+* By default, `cylc install` will install the workflow found in `$PWD` into
+  `~/cylc-run/$(basename $PWD)/runN` (where `runN` = `run1`, `run2` ...).
+* If `run1` already exists it will install into `run2` (and so on).
+  Create a symlink `runN` pointing at the latest run.
```
wxtim, line 108: Might `run-last` be more intuitive?

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,228 @@ The remaining points can happen any time after 1) is complete although note that
...
+* `--run-name=my-run` implies install into `~/cylc-run/$(basename $PWD)/my-run`.
```
wxtim, line 109: Possible enhancement suggestion:
`--run-name-pattern=my-run-%(n)d` implies install similar logic to default, but with specified alternative to `runN`

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,228 @@ The remaining points can happen any time after 1) is complete although note that
...
+  If the target directory already exists then fail.
+* `--no-run-name` implies install into `~/cylc-run/$(basename $PWD)`.
+* `--flow-name=my-flow` implies install into `~/cylc-run/my-flow/runN`.
+* `--directory=/path/to/flow` (`-C ...`) implies install the workflow found in
```
wxtim, line 113: consider aliases `--source-dir`, `--src-dir`?

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,228 @@ The remaining points can happen any time after 1) is complete although note that
...
+    e.g. you cannot install into `~/cylc-run/my-flow/runN` if `~/cylc-run/my-flow`
+    already contains an installed suite.
+* Version control information will be recorded where relevant (svn/git).
+* The `log, `share`, share/cycle` and `work` directories  will be created
+  following whatever symlink rules are defined for localhost
+  (over-ridable via command line options).
+  * This will continue to support "symlinked alternate run directories"
+    (see [cylc-flow #2935](https://github.com/cylc/cylc-flow/pull/2935)).
+
+`cylc play` - new command to run an installed workflow
```
wxtim, line 136: @oliver-sanders and I should get inkscaping when it comes to the docs for this! The cassette player conceit has real potential for helping users understand what is going on and we should exploit that mercilessly.

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,228 @@ The remaining points can happen any time after 1) is complete although note that
...
+  the installed workflow, not in the source directory).
+* The rsync will use `--delete` to ensure that any previously installed files
+  which have been removed from the source directory also get removed from the
+  installed workflow.
+  * Note that this will have the effect of removing any files installed via
+    `rose-suite.conf`. They will get recreated by a subsequent reload or play but
+    could affect a running workflow.
+* Any options specified as part of the original install will be retained unless
+  overridden on the command line.
+* Consider a `--dry-run` option to report what would be changed?
```
wxtim, line 170: Can we get this script to store a record of changes somewhere so that when people mess their workflows up we can see what they've done?

wxtim:
I feel that this is well enough defined that I'd be happy to get cracking on it, which suggests to me that the detail level is pretty good.
We're going to have a massive project documenting all this.

wxtim (approved):

docs/proposal-rose-suite-run.md
oliver-sanders: Should probs make this pattern a global config where a blank value implies don't use a run dir?

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,228 @@ The remaining points can happen any time after 1) is complete although note that
...
-   * Currently this functionality assumes a shared filesystem but we should
-     support workflow server hosts with separate filesystems (future
-     enhancement?).
-   * Needs to create the run time directory structure which has been configured
-     for the `localhost` platform.
+2. Support installation of a suite from the source location (working copy) to
+   the run location (`cylc-run`).
+   * This functionality assumes the source location (working copy) and the run
+     location (`cylc-run`) are both accessible on the host where the command is
+     run.
```
oliver-sanders, line 49: Reinforced by the creation of a `source` symlink in the installed workflow (as presently generated during registration) coupling the installation and source.

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,228 @@ The remaining points can happen any time after 1) is complete although note that
...
      separate step. This was difficult previously because cylc loads the suite
      after it has been detached?
-   * Consider making all validation strict? If we want really want to allow
-     naked dummy tasks in some circumstances then we would be better to add a
-     suite configuration for this.
+   * Consider making all validation strict?
+     * If we really want to allow naked dummy tasks in some circumstances then
+       we would be better to add a suite configuration for this?
+     * Note: we don't allow you to inherit from an undeclared ("naked") family
+       so why do we allow you to run an undeclared task?
```
oliver-sanders, line 69:
```suggestion
     * If we really want to allow naked tasks in some circumstances then
       we would be better to add a suite configuration for this?
```

A facetious correction, however, naked tasks are not dummy tasks, they still run and just because they are not explicitly defined does not mean they haven't been configured.

Consider this common pattern:

```ini
[runtime]
    [[root]]
        script = rose task-run
```

So-called "naked" tasks are a valid feature of Cylc. There can be cases where explicitly defining runtime tasks may be inconvenient, e.g. for highly programatic suites where obtaining a full list of tasks present in the graph may require building dependencies in Python and writing them in Jinja2 (the other option being introspection 🤮).

I feel that this is somewhat over-opinionated Rose logic I would like to loose. Propose printing a warning listing the first `N` naked tasks. Note that this kind of check would normally be performed by a static code analyser and not provide a barrier to execution.

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,228 @@ The remaining points can happen any time after 1) is complete although note that
...
+   * This is big change in working practise but Cylc 8 is a good time to
+     introduce it.
    * Use a symbolic link to identify the latest run which is the default name
      when performing operations on a suite?
 
-8. Consider the behaviour created by the `CYLC_VERSION` and `ROSE_VERSION`
-   variables in `rose-suite.conf`. These allow rose suite-run to:
-   * Change the version of Cylc and check the version of Rose used to run the
-     suite.
-   * Prevent suites accidentally upgrading or downgrading versions on restart.
```
oliver-sanders, line 88: For reference, here's a bit of context as to why this functionality has been removed:

We would expect that Rose would be installed in the same environment as Cylc so offering the option to configure the Rose version independently of the Cylc version does not make sense.

Note, however, that Rose will need to continue to use a wrapper script approach for locating remote installations as the Cylc environment is not activated in [remote and batch] task environments. We may decide to support this via a `ROSE_VERSION` environment variable as at present, though, since the Rose-Cylc couplings should be consistent across all installations, we could potentially rely on the `CYLC_VERSION` variable for the same functionality.

Perhaps one day version limiting could be done as part of https://github.com/cylc/cylc-flow/issues/3780?

docs/proposal-rose-suite-run.md
oliver-sanders: (preferably we would use some form of abstraction to provide support for others as well)

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,228 @@ The remaining points can happen any time after 1) is complete although note that
...
+   * `rose suite-run` current supports this for Subversion only
+     (`svn info` + `svn status` + `svn diff`).
+   * Extend this to support equivalent information for git.
+
+### CLI changes
+
+`cylc install` - new command to install workflows.
+* By default, `cylc install` will install the workflow found in `$PWD` into
+  `~/cylc-run/$(basename $PWD)/runN` (where `runN` = `run1`, `run2` ...).
+* If `run1` already exists it will install into `run2` (and so on).
```
oliver-sanders, line 107:
```suggestion
* If `--run-dir` is not specified and `run1` already exists it will install into `run2` (and so on).
```

docs/proposal-rose-suite-run.md
oliver-sanders: I think this is ok.

This option effectively specifies the path to a Cylc configuration file/dir so follows the pattern of commands like `make` which uses `--directory`, `-C`.

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,228 @@ The remaining points can happen any time after 1) is complete although note that
...
+* `--no-run-name` implies install into `~/cylc-run/$(basename $PWD)`.
+* `--flow-name=my-flow` implies install into `~/cylc-run/my-flow/runN`.
+* `--directory=/path/to/flow` (`-C ...`) implies install the workflow found in
+  `/path/to/flow` (rather than `$PWD`).
+* All the options accepted by `cylc play` will be accepted by `cylc install`.
+  These options will provide the defaults for when `cylc play` is run with this
+  installed workflow (but they can still be overridden). Relevant options will
+  also apply to other cylc commands when operated on the installed workflow
+  (e.g. `cylc validate`). Note that supporting this will require the creation of
+  a new options file.
```
oliver-sanders, line 120: I'm not convinced by this "install now with some options then run later" use case. Does anyone actually want to do this?

It is currently possible to "install with suite variables but not run" in Rose (`rose suite-run -l`) but this is seldom done. The only use case (beyond evaluating the result of running `rose suite-run`) that I'm aware of is for running "static analysis" commands like `validate` and `get-config`. With the retirement of `rose suite-run` these commands can be run directly against the source directory negating the need to install the workflow.

Note that we have talked about implementing an options file for Cylc in Python as an alternative to the `rose suite-conf` for specifying suite variables, etc (at Cylc workshops). So if we push ahead with this I would suggest approaching it from that angle. This file would provide good transparency to users (as it would be user-facing and could be defined in the src_dir) as well as supporting `src_dir == run_dir` uses.

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,228 @@ The remaining points can happen any time after 1) is complete although note that
...
+* Installation will involved copying over the files found in the source
+  directory.
+  * This will exclude any `.git`, `.svn`, `log`, share` or `work` directories.
```
oliver-sanders, line 123: Just wondering if users might want to copy over `.git` directories. Perhaps they are using git sub-modules that they want to install in the run dir or something like that.

If needed could make this configurable via a `.cylcignore` file (default to the above, if present add `log` to the user-defined list).

docs/proposal-rose-suite-run.md
dpmatthews: Done

docs/proposal-rose-suite-run.md
dpmatthews: Now noted in proposal

docs/proposal-rose-suite-run.md
dpmatthews: Don't really mind. I suggested `runN` because we currently use `NN` as a link to the latest job in the log directory.

docs/proposal-rose-suite-run.md
dpmatthews: I'm not sure a global config is a good idea since it will cause `cylc install` to behave differently which makes support more difficult - they can always set up a command alias.

I'd suggest leaving any enhancement until we see whether there is demand (it wouldn't require any new command line option).

docs/proposal-rose-suite-run.md
dpmatthews: Personally I think there are very few cases where you want to run a task which is purely defined by the root environment. If we allow workflows to run with naked dummy tasks (something `rose suite-run` prevents) I fear we will end up with lots more mistakes and lots of wasted time / resources.

docs/proposal-rose-suite-run.md
dpmatthews: Done

docs/proposal-rose-suite-run.md
dpmatthews: OK. I've removed this bit for the moment (we can always change our minds later).
I won't resolve it for the moment in case anyone else wants to comment.

docs/proposal-rose-suite-run.md
dpmatthews: Added to proposal

docs/proposal-rose-suite-run.md
hjoliver: > A facetious correction, however, naked tasks are not dummy tasks, they still run and just because they are not explicitly defined does not mean they haven't been configured.

Fair point but the word "dummy" is somewhat semantically fraught.  I've always meant "dummy task" as a real task that has had its guts replaced by `/usr/bin/true` or whatever.  As opposed to a "simulated task" which is what you get in simulation mode. A "naked dummy task" is one that had no guts defined in the first place, but is given a `/usr/bin/true`.

hjoliver: I think the one good use for naked dummy tasks is for experimenting with Cylc scheduling (in any number of ways) when you don't care about the task content.  Just write some graph structure and run to see if it behaves as you expect or whatever.  This would be significantly (although not horrifically) more painful if we were forced to write a `[runtime]` subsection for every task.  I'd be open to requiring an explicit configuration setting to allow it though (`allow naked dummy tasks = True` or whatever).  Also open to considering a better name for these things, if we want to go there.

hjoliver: The new config setting could come with a validation warning: "WARNING allowing naked dummy tasks is dangerous, please disallow it before running this flow in production" or something.

docs/proposal-rose-suite-run.md
hjoliver: Not sure I fully understand this.
> the Cylc environment is not activated in [remote and batch] task environments

It will be activated in task job environments, to provide access to `cylc message` and other `cylc` commands, no?

docs/proposal-rose-suite-run.md
hjoliver: IMO `last` is somewhat ambiguous - it could mean this is the final (last ever) run, rather than just the lastest run in an ongoing series of runs.

I like `runN` because it evokes the well known (?) `1, 2, ..., N` representation of an integer series (`run1, run2, ..., runN`)

I suppose `run1, run2, ..., run-latest` might be more intuitively obvious to users, but it's a bit ugly.

docs/proposal-rose-suite-run.md
hjoliver: `--run-name-pattern=road-experiment-%(n)d` I'm not sure this sort of global replacement of `run%d` would add much.  Wouldn't it be better to name the flow instead @wxtim ?  (So, `road-experiment/run1` etc.)

hjoliver: > Should probs make this pattern a global config where a blank value implies don't use a run dir?

What do you mean by "don't use a run dir" @oliver-sanders ? Replicate the current model of overwriting the existing one every time?

hjoliver: (Ah, `--no-run-name` just below might have answered my question)

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,237 @@ The remaining points can happen any time after 1) is complete although note that
...
+* `--no-run-name` implies install into `~/cylc-run/$(basename $PWD)`.
+* `--flow-name=my-flow` implies install into `~/cylc-run/my-flow/runN`.
+* `--directory=/path/to/flow` (`-C ...`) implies install the workflow found in
+  `/path/to/flow` (rather than `$PWD`).
+* Installation will involved copying over the files found in the source
+  directory.
+  * This will exclude any `.git`, `.svn`, `log`, share` or `work` directories.
+    Everything else will be copied.
+    * If needed we could make this configurable via a `.cylcignore` file (future
+      enhancement). Note that `log`, share` & `work` would always be excluded.
```
hjoliver, line 122: (Why would there be `log`, `share`, and `work` in the source directory?)

hjoliver: (Maybe installation should abort if those are found, as it would suggest a run directory not a source directory)

docs/proposal-rose-suite-run.md
hjoliver: Yes, and ruthlessly too!

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,237 @@ The remaining points can happen any time after 1) is complete although note that
...
+`cylc reload` - no change from existing command.
+
+`cylc validate` - as per existing command with the following changes:
+* If validating a workflow installed with `cylc install` then use any relevant
+  command line options specified as part of the install unless overridden on the
+  command line.
+* Will work fine on source directories which include `rose-suite.conf` files.
+* No `--strict` mode
+  (naked dummy tasks no longer allowed - all tasks must be declared).
+  * This one is controversial and requires further discussion!
```
hjoliver, line 187: (See comments above!)

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,237 @@ The remaining points can happen any time after 1) is complete although note that
...
+* Cyclic graph validation can be quite slow and is currently included as part of
+  `--strict`. Add a `--no-cyclic-graph-validation` option to disable this?
```
hjoliver, line 189: Or should cyclic graph detection be opt-in?

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,237 @@ The remaining points can happen any time after 1) is complete although note that
...
+* `cylc reinstall-reload`
+
+`cylc pause|unpause` - new commands replacing `cylc hold|release|unhold`
+* Note that when you use `cylc stop` the scheduler will be modified to unpause
+  the workflow before stopping.
+  This means that a stopped workflow will not be paused when it is continued via
+  `cylc play`. On the other, a workflow which is continued for any other reason
+  (e.g. the server died) will retain its paused status.
+
+`cylc register` - remove: functionality no longer required.
```
hjoliver, line 204: :open_mouth: holy sh@t, that the end of an era innit.

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,237 @@ The remaining points can happen any time after 1) is complete although note that
...
+  For example, if you clean the `log` directory this allows you to rerun a
+  workflow (keeping the existing data in `share` if that is what you want).
+  The default will be to remove everything including the top level directory.
+* Support an option to clean all install targets other than localhost?
+* We may also want an option to archive the log directory rather than remove it.
+  This could simply imply renaming the directory, perhaps with a timestamp
+  (similar to what `rose suite-run` does but without tar gzipping the
+  directory). This would allow a workflow to be warm started without completely
+  losing the previous log files (this is an operational requirement at present
+  although we hope that re-flows will reduce or remove the need for this).
```
hjoliver, line 221: Ooh, a use case for reflows :boom:

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,237 @@ The remaining points can happen any time after 1) is complete although note that
...
+  Note that we would not provide any further support for these archived log
+  directories (e.g. no support for accessing the logs via the UI).
+* The file `log/rose-suite-run.locs` is currently required by `rose suite-clean`
+  but `cylc clean` should probably get this information from the database?
+
+Note that many commands currently have REG / SUITE as a required argument
+(e.g. `cylc validate`). Where possible these will be changed to be optional
+arguments with the default being `$PWD`.
```
hjoliver, line 229: Good, I've always hated that use of `REG` :+1:

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,237 @@ The remaining points can happen any time after 1) is complete although note that
...
+* We could also consider using an option rather than an argument.
+
+For all commands which have been replaced (run, restart, hold, register,etc) we
+will replace them with (hidden, not listed in CLI help) commands which simply
+report what commands should be used in their place.
+
+Once the main CLI changes are in place we should do a complete review of the
+entire cylc command set to see whether there are further commands which should
+be retired or altered.
```
hjoliver, line 238: Yes :+1:

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,237 @@ The remaining points can happen any time after 1) is complete although note that
...
+  * `rose suite-shutdown`
+* As with cylc, we will provide hidden commands which simply report what
+  commands should be used in their place.
+* Assume we will still need to support `rose stem`.
+
+Related issues / notes:
+* [cylc-flow #1030](https://github.com/cylc/cylc-flow/issues/1030)
+* [Workshop notes](https://cylc.github.io/cylc-admin/feb2020-workshop-notes#wednesday).
+  Note that this proposal differs in a number of ways from what was discussed at the workshop:
+  * no git version control of run dir (future enhancement?)
```
hjoliver, line 259: We should definitely consider it in future, but not a must-have for 8.0 :+1:

hjoliver (commented):
Very good :tada:  - happy to approve this quickly once the few discussion points are resolved.

docs/proposal-rose-suite-run.md
oliver-sanders: If we make the run name pattern user-definable then it might be best to use `NN` or something suitably neutral.

docs/proposal-rose-suite-run.md
oliver-sanders: > I'd suggest leaving any enhancement until we see whether there is demand (it wouldn't require any new command line option).

Fair enough, happy with that, my main motive in posting it was to make it easy to disable the run dir hier.

docs/proposal-rose-suite-run.md
oliver-sanders: > I think the one good use for naked dummy tasks is for experimenting with Cylc scheduling

Like for instance, the Cylc Tutorial!

(Users hit more issues in the scheduling section than the runtime section since all of the workflow complexity is there so it makes sense to tackle the graph first.)

docs/proposal-rose-suite-run.md
oliver-sanders: > It will be activated in task job environments, to provide access to cylc message and other cylc commands, no?

NO!!!!!!!

Cylc commands can be called via the wrapper script, activating the environment would change the Python interpreter and the entire Cylc stack.

At present the Cylc environment is only (accidentally) activated in local background/at jobs but that's going to change - https://github.com/cylc/cylc-flow/pull/3712

oliver-sanders: Just thinking about this for a moment, I think we may require an additional `isodatetime` wrapper script if we wish to migrate users from the deprecated `rose date` and `cylc cycle-point` commands.

docs/proposal-rose-suite-run.md
oliver-sanders: > (Why would there be log, share, and work in the source directory?)

There shouldn't be! This change makes it an error to have any of these directories in the src dir to avoid the nasty results of rsync'ing one over the other.

docs/proposal-rose-suite-run.md
oliver-sanders: Happy with that, we could also turn it on automatically if the number of tasks is less than, say 100, else print an info line with the `--cyclic-graph-validation` CLI opt to raise awareness.

docs/proposal-rose-suite-run.md
oliver-sanders: TBH we've really just renamed `cylc register` to `cylc install`, the same functionality is there :/

docs/proposal-rose-suite-run.md
oliver-sanders: And quite a nice one too as warm starts aren't the nicest. This would require the ability to stop the previous flow.

docs/proposal-rose-suite-run.md
oliver-sanders: Note that we still have the use case of `cylc stop <name>` and the desired functionality of `cylc stop <pattern>`, `cylc clean <pattern>`, etc.

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,237 @@ The remaining points can happen any time after 1) is complete although note that
...
+* No `--strict` mode
+  (naked dummy tasks no longer allowed - all tasks must be declared).
+  * This one is controversial and requires further discussion!
+* Cyclic graph validation can be quite slow and is currently included as part of
+  `--strict`. Add a `--no-cyclic-graph-validation` option to disable this?
+
+Several combinations of the above commands will be commonly used.
+Propose to support these as separate commands:
+* `cylc install-play` (`cylc ip` for short?)
+* `cylc install-validate`
```
oliver-sanders, line 194: I think this use-case is obsolete now as `cylc play` will perform validation anyway (`cylc run` already performs 95% of validation) and with the Rose plugin providing support for the `rose-suite.conf` file we will be able to run `cylc validate` directly on the source directory.

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,237 @@ The remaining points can happen any time after 1) is complete although note that
...
+  command line.
+* Will work fine on source directories which include `rose-suite.conf` files.
+* No `--strict` mode
+  (naked dummy tasks no longer allowed - all tasks must be declared).
+  * This one is controversial and requires further discussion!
+* Cyclic graph validation can be quite slow and is currently included as part of
+  `--strict`. Add a `--no-cyclic-graph-validation` option to disable this?
+
+Several combinations of the above commands will be commonly used.
+Propose to support these as separate commands:
```
oliver-sanders, line 192: We may want to consider what the different options for combining these commands is since there are different models we could adopt.

Suggest not worrying too much about this now and awaiting user feedback. In the mean time there is something rather therapeutic about calling `install` then `play` as it reinforces understanding of the behaviours of the commands.

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,237 @@ The remaining points can happen any time after 1) is complete although note that
...
+* `cylc install-play` (`cylc ip` for short?)
+* `cylc install-validate`
+* `cylc reinstall-reload`
+
+`cylc pause|unpause` - new commands replacing `cylc hold|release|unhold`
```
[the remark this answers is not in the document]
oliver-sanders: @dpmatthews is not so fond of this command combination.

I'm rather fond of keeping functionalities separate but admit-ably do appreciate the simplicity of the tape-deck model and the removal of another major command.

docs/proposal-rose-suite-run.md
hjoliver: >  (In the mean time there is something rather therapeutic about calling install then play as it reinforces understanding of the behaviours of the commands.)

:+1:

docs/proposal-rose-suite-run.md
hjoliver: Why should we make it user-definable though? (see next discussion below - surely the metadata conveyed by a less generic name for "run", which is the same for every run!, should be one level up, in the flow name?

docs/proposal-rose-suite-run.md
oliver-sanders: Putting it into a new config would kinda kill this use by making it clunky.

docs/proposal-rose-suite-run.md
hjoliver: I'm not sure why we should allow disabling the new run1, run2, ... structure either (if that's what you mean). Is there some valid case where continually overwriting previous run-dirs is actually desirable?

docs/proposal-rose-suite-run.md
hjoliver: So, maybe it is sufficient to allow naked dummy tasks (because sometimes it is very convenient) but WARN if any are detected that it is a dangerous practice in production mode.

docs/proposal-rose-suite-run.md
hjoliver: > NO!!!!!!!
> 
> Cylc commands can be called via the wrapper script, activating the environment would change the Python interpreter and the entire Cylc stack

He heh, sorry, I wasn't thinking very deeply there - that is what I meant (the Cylc environment is activated in the job script, by the `cylc` wrapper, whenever a Cylc command is called) - but yeah that doesn't help separate use of the Rose CLI.

docs/proposal-rose-suite-run.md
hjoliver: I'm fine with this. The only use case I'm aware of is to install then validate, with `rose suite-run`.

docs/proposal-rose-suite-run.md
oliver-sanders: Course not, this just comes out of the fear that users might find run1, run2, ... somewhat prescriptive.

Ultimately it's just a registration pattern, it's not a hard-baked Cylc feature.

docs/proposal-rose-suite-run.md
hjoliver: > Note that `log`, share` & `work` would always be excluded.

This says those directories will be excluded, but it doesn't make it an error for them to exist.  IMO it should be and error. (Or have I misunderstood?)

docs/proposal-rose-suite-run.md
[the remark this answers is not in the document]
hjoliver: > the removal of another major command.

Let's do it then. We have too many commands!

docs/proposal-rose-suite-run.md
hjoliver: Meh, `cylc install` does a lot more!

hjoliver: (And "install" is a more intuitive concept too, for users, IMO).

docs/proposal-rose-suite-run.md
hjoliver: We have that ability: `cylc stop --flow=LABEL` :boom:

docs/proposal-rose-suite-run.md
oliver-sanders: Ah, sorry, I'm getting conversations mixed up with the proposal.

Yes, this would silently ignore these dirs. I agree this should be an error or at the every least a warning.

docs/proposal-rose-suite-run.md
hjoliver: (My point there was just that I never liked - although it was probably my fault! - use of the unintuitive "REG" metavar.  It should have been "SUITE-NAME" or similar.)

docs/proposal-rose-suite-run.md
oliver-sanders: T'would be my preference.

Users need to get more adept at reading warnings. We need to get better at simplifying output to encourage this practice.

docs/proposal-rose-suite-run.md
dpmatthews: > Just thinking about this for a moment, I think we may require an additional `isodatetime` wrapper script if we wish to migrate users from the deprecated `rose date` and `cylc cycle-point` commands.

I fear you're right - yuck!

docs/proposal-rose-suite-run.md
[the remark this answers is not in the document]
dpmatthews: I'm confused. Are you suggesting we use the play command to unpause tasks?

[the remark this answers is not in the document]
hjoliver: Yes, that's what I'm suggesting, unless there's a downside beyond what I've suggested in the first comment above (or you think that is in itself sufficiently bad).

docs/proposal-rose-suite-run.md
hjoliver: I think "play around using dummy tasks" undersells something that we should encourage users to do a lot more of - instead of trying to get some new (to them) aspect of Cylc working by experimenting with their real suites (which is slow and hard to debug etc.). 

But fair enough; new Issue: https://github.com/cylc/cylc-flow/issues/3866

docs/proposal-rose-suite-run.md
dpmatthews: I suggest we leave the proposal as is - we can always come back to this idea later

docs/proposal-rose-suite-run.md
dpmatthews: Issue created: cylc/cylc-flow#3869

docs/proposal-rose-suite-run.md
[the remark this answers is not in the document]
dpmatthews: The interface for `cylc release` is completely different, e.g.:
 cylc release REG TASK_GLOB ... - release one or more tasks
It doesn't make sense to me to make this part of `cylc play`.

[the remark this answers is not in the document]
dpmatthews: I agree `unpause` is ugly but I'm not sure what else to suggest.
I'd be happy to stick with hold/release, especially since we still plan to use `held` rather than `paused` as a task attribute:
https://cylc.github.io/cylc-admin/proposal-state-names.html#new-abstract-task-states

docs/proposal-rose-suite-run.md
dpmatthews: Proposal updated to make this an error

docs/proposal-rose-suite-run.md
dpmatthews: OK - removed

docs/proposal-rose-suite-run.md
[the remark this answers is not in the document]
hjoliver: Sorry, you're right, I forgot about holding individual tasks. Although we still need to figure out how to do that under SoD  https://github.com/cylc/cylc-flow/issues/3743

docs/proposal-rose-suite-run.md
hjoliver: https://github.com/cylc/cylc-flow/issues/3864#issuecomment-708699402

docs/proposal-rose-suite-run.md
hjoliver: https://github.com/cylc/cylc-flow/issues/3871

dpmatthews:
@oliver-sanders @hjoliver - is this good enough to go in now?

docs/proposal-rose-suite-run.md
oliver-sanders: (Commented on the Cylc wrapper script PR suggesting using `CYLC_VERSION` for the `isodatetime` wrapper script - https://github.com/cylc/cylc-flow/pull/3754#issuecomment-709031881)

docs/proposal-rose-suite-run.md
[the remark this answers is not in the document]
oliver-sanders: Once the held state is implemented via xtriggers we will require a `set|satisfy xtrigger` interface (which would be quite nice, say for adding/removing clock triggers via the GUI). So holding a task becomes a special case of that.

* play/pause for suite state, hold/release for task state?
* pause/release for both?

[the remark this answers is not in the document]
dpmatthews: ... or hold/release for both? (i.e. no change)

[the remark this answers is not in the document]
hjoliver: If we're going for the tape deck analogy, we might as well be consistent, which suggests we need play/pause at least as an alias.  As per Oliver's comment, play/pause makes sense for the flow as a whole, but not really for individual tasks.

So my vote is:
- play/pause for flow state, with aliases: play=release=unpause (if paused) and pause=hold
- hold/release for task state, with aliases: hold=pause and release=unpause

(The aliases are all pretty obvious and easily understood, but will help out users if the flow vs task distinction bleeds together a bit in the minds...)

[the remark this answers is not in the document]
dpmatthews: I've updated the proposal to add the aliases.
The use of `cylc play` to release a paused workflow is already noted as an option.
Could we get this merged and make a decision on that later?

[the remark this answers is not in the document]
oliver-sanders: Ok, add a TODO note or something to mark this apart from the more solid parts of the proposal.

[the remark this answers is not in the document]
dpmatthews: Done

docs/proposal-rose-suite-run.md
```diff
@@ -39,66 +39,234 @@ The remaining points can happen any time after 1) is complete although note that
...
+  * The installation should fail if neither `suite.rc` nor `cylc.flow` exist.
+  * The installation should fail if the target directory is not valid.
+    e.g. you cannot install into `~/cylc-run/my-flow/runN` if `~/cylc-run/my-flow`
+    already contains an installed suite.
+* Version control information will be recorded where relevant (svn/git).
+* The `log, `share`, share/cycle` and `work` directories  will be created
+  following whatever symlink rules are defined for localhost
+  (over-ridable via command line options).
+  * This will continue to support "symlinked alternate run directories"
+    (see [cylc-flow #2935](https://github.com/cylc/cylc-flow/pull/2935)).
```
oliver-sanders, line 131: Missed this on my first scan, the symlink functionality will allow the entire run dir to be symlinked to another location which would effectively replace this functionality?

```suggestion
  * This will replace "symlinked alternate run directories" functionalty
    (see [cylc-flow #2935](https://github.com/cylc/cylc-flow/pull/2935)).
```

May be desirable to implement a replacement CLI option though?

dpmatthews: Change applied.
I've haven't tried to  define what command line options will be used - we can decide this later.

oliver-sanders (approved):
Happy enough to approve this as being good enough to get on with. There are some areas that are yet to be agreed but these are minor in the scheme of the works to be undertaken and our thinking may well evolve during implementation (as it usually does!).

hjoliver (approved):
:tada:

PR #112, opened 2020-10-14
Add n-distance window proposal

hjoliver:
Tried to write up the outcomes of several recent discussions about n-distance windowing here, to make sure we're all aware and agreed. 

The main concept and some details have already been accepted of course, but several sub-issues had not been thought through yet.  I've written it up from main concept down though for completeness, so we can extract documentation in future of the window users will see in the Cylc 8 UI.

The main new thing here is: how to handle satisfied tasks that are held back by xtriggers or queues etc. - these are not "active" (n=0) and they're not n=1 relative to current active tasks either, but they should not be invisible. I've argued here that the right thing to do is artificially add them to the n=1 window.

Added the main graph-window suspects as reviewers.

### DIFF
--- a/docs/proposal-n-distance-window.md
+++ b/docs/proposal-n-distance-window.md
@@ -13,7 +13,7 @@ This is really the only well-founded, easily comprehended, and scalable way to
 display large Cylc-style cycling workflows. Only in smaller cases is it
 reasonable to show all the tasks in active cycles.
 
-## The Datastores
+## Datastores
 
 Windowing will be managed by similar datastores in the scheduler and UI Server
 - the scheduler only needs `n = 0` for scheduling
@@ -30,7 +30,7 @@ PROPOSE: stop at `n = 1` in the future direction
 
 ## Disappearing waiting tasks?
 
-### non-task prerequisites
+### 1. non-task prerequisites
 
 Tasks may be held back from becoming active even though their graph/task
 prerequisites are all satisfied, by runahead limit, queues, xtriggers,
@@ -61,7 +61,7 @@ foo => bar
 @queue => bar
 ```
 
-### parent gaps
+### 2. parent gaps
 
 A task with multiple parents will look like it is "coming next" (`n = 1`) with
 respect to each active parent, even though that won't really be true until 
@@ -76,7 +76,7 @@ active, `C` will be visible in `n = 1` when `A` is active, then it will
 disappear from the UI in the gap between `A` and `B`, and finally appear again
 when `B` is active.
  
-### Possible solutions
+**Solutions?**
 
 1. live with it? It probably won't happen very often, and it's easy to explain
      to users (strict `n = 1` edge from active tasks)
--- a/docs/proposal-n-distance-window.md
+++ b/docs/proposal-n-distance-window.md
@@ -1,0 +1,100 @@
+# Cylc 8 n-distance windows for the UI
+
+Cylc 8 spawn-on-demand makes it possible to display (to users) an intuitive
+graph-based window centered on current active tasks.
+
+## Basic Concept
+
+*The UI, in all views, should display current active tasks plus those out to
+`n` graph edges from the active ones. So `n = 0` refers to active tasks;
+`n = 1` to the parents and children of the active tasks; and so on.*
+
+This is really the only well-founded, easily comprehended, and scalable way to
+display large Cylc-style cycling workflows. Only in smaller cases is it
+reasonable to show all the tasks in active cycles.
+
+## The Datastores
+
+Windowing will be managed by similar datastores in the scheduler and UI Server
+- the scheduler only needs `n = 0` for scheduling
+  - but its datastore may go out to `n = 1` for convenience
+- the UIS will compute higher-n windows and populate non-active tasks from DB
+  - to do that it will need the graph pattern from the scheduler
+
+## Future Tasks
+
+PROPOSE: stop at `n = 1` in the future direction
+- `n = 1` shows what comes next in the flow; beyond that nothing has happened
+  yet and showing it will unnecessarily reduce screen space for active tasks
+- (plus, Cylc 9 changes may make the future graph less predictable)
+
+## Disappearing waiting tasks?
+
+### non-task prerequisites
+
+Tasks may be held back from becoming active even though their graph/task
+prerequisites are all satisfied, by runahead limit, queues, xtriggers,
+and task hold. These will not be visible in `n = 0` because they're not
+active, or in `n = 1` because their parents are no longer active. Users will
+see `n = 1` tasks (which are supposed to show what's coming next) disappear
+from the UI, and suddenly reappear later when their queue (say) releases them.
+
+*This problem is an artifact of the fact that the graph only represents task
+dependencies*. Xtriggers (etc.) should really be represented as entities in the
+graph that persist at `n = 0` (c.f. long-running tasks that periodically emit
+outputs to satisfy children), then the problem tasks can be seen as just normal
+`n = 1` tasks that are still waiting on their active xtrigger (etc.) parents.
+They would stay in `n = 1` until *all* prererquisites (not just task
+prerequisites) are satisfied, then go straight to `n = 0` as normal. In light
+of this:
+
+PROPOSE: keep these `waiting` tasks visible in `n = 1` until active
+- (add artificially to `n = 1` as edge-following from `n = 0` won't find them).
+
+POST CYLC 8:
+- figure out how to represent all dependence in the graph
+- unify task and non-task prerequisites?
+- perhaps take an event-driven approach to queues etc. too?
+
+```
+foo => bar
+@queue => bar
+```
+
+### parent gaps
+
+A task with multiple parents will look like it is "coming next" (`n = 1`) with
+respect to each active parent, even though that won't really be true until 
+the very last of its parents becomes active. That's fine if all the parents run
+concurrently, but otherwise the child can repeatedly disappear and reappear
+until it becomes active itself. This will likely be disconcerting to users.
+```
+A & B => C
+```
+Here, if there is a significant gap between `A` succeeding and `B` becoming
+active, `C` will be visible in `n = 1` when `A` is active, then it will
+disappear from the UI in the gap between `A` and `B`, and finally appear again
+when `B` is active.
+ 
+### Possible solutions
+
+1. live with it? It probably won't happen very often, and it's easy to explain
+     to users (strict `n = 1` edge from active tasks)
+2. artificially exclude the child from `n = 1` until the last parent is active?
+     - check for other unsatisfied prerequisites before adding the child?
+3. always show all parents of a child at `n = 1`? 
+     - the child will still disappear in a parent gap, but at least that will
+       be foreshadowed by visible inactive parents before it does disappear
+
+PROPOSE: implement (1) initially, and later consider (2) and (3) as
+enhancements. (2) looks best but it may be tricky avoid wrongly excluding a
+child from `n = 1` in alternate path graphs?
+
+## Graph Isolates
+
+Windowing will often result in multiple disconnected graphs in the graph view
+(even without considering reflows). We will just have to live with that
+(because, see Justification above).
+
+POST CYLC 8: consider ways to display multiple disconnected graphs better, such
+as by bridging with "scissor nodes" as in the Cylc 7 GUI?


docs/proposal-n-distance-window.md
```diff
@@ -0,0 +1,100 @@
...
+ 
+### Possible solutions
+
+1. live with it? It probably won't happen very often, and it's easy to explain
+     to users (strict `n = 1` edge from active tasks)
+2. artificially exclude the child from `n = 1` until the last parent is active?
+     - check for other unsatisfied prerequisites before adding the child?
+3. always show all parents of a child at `n = 1`? 
+     - the child will still disappear in a parent gap, but at least that will
+       be foreshadowed by visible inactive parents before it does disappear
```
dpmatthews, line 87: This has got me thinking about other cases.

```
A | B => C
```
In this case you'd want C in `n = 1` if either A or B are active.
But what if A is active but then fails (handled by a failure trigger)?
C is then going to disappear from "n=1" until B is active even with options 1 or 2.

```
A:fail & B:fail => C
```
What happens if A succeeds?
Will C appear in "n=1" when B is active even though it can never run?

```
get_obs<station> => process_obs
```
(where "station" is a large list)
With options 1 or 2, as each "get_obs<station>" task succeeds it will immediately disappear from "n=1".
Then, when "process_obs" eventually runs, all the "get_obs<station>" tasks will reappear in "n=1".
Option 3 would definitely be better in this case?

hjoliver: > In this case you'd want C in n = 1 if either A or B are active.
But what if A is active but then fails (handled by a failure trigger)?
C is then going to disappear from "n=1" until B is active even with options 1 or 2.

That's OK IMO because `A => C` says `C ` is coming next only if `A` succeeds.  If `A` fails, then `C` is not coming next any more and can disappear.  If `B` comes along later, then `C` appears again, but we can't keep `C` visible in anticipation of that because its possible that `B` will never become active.

hjoliver: > What happens if A succeeds?
Will C appear in "n=1" when B is active even though it can never run?

I think this case is OK too.  `n = 1` coming true is always going to be predicated on the right uptream output actually happening.

hjoliver: > With options 1 or 2, as each "get_obs" task succeeds it will immediately disappear from "n=1".

Presumably you meant "n=0"? (each `get_obs` task is in `n=0` when it is running).

There is no problem here if all the parents (get_obs<station>) run concurrently, even if they all finish at different times.  `process_obs` will only disappear if there's a gap with none of them running. So this isn't really any different than my `A & B => C` example is it? (Just more parents).

hjoliver: Seems to me we should go with the "live with it" option and see how it looks in the UI.  Maybe we just educate users that `n=1` doesn't necessarily mean "coming next" (it may depend on other tasks too, that are not currently active).

hjoliver: Note that option 3. won't stop the disappearing in a "parent gap". It will show all the parents (instead of just the active ones) then they will all disappear in the gap.

dpmatthews: > Presumably you meant "n=0"? (each `get_obs` task is in `n=0` when it is running).

Sorry, not well explained. I'm trying to think through what happens when viewing this workflow with n=1.
I could have, say, 50 different `get_obs<>` tasks running.
As each of these tasks completes it will immediately disappear from the display which isn't ideal (users don't get to see the completion status of the task).
Eventually you'll be left with 1 `get_obs<>` task shown in the display plus the `process_obs` task.
When that final `get_obs<>` task completes, `process_obs` will start running and suddenly all 50 `get_obs<>` tasks will re-appear in the display - again, not ideal.

hjoliver: Yeah a window based strictly on graph distance from active tasks is always going to exhibit this sort of behaviour. I suppose your 50-member family (or similar) example is an extreme case of it. 

Another idea: treat partially satisfied waiting tasks as n=1 instead of hiding them.  Then in your example above, `get_obs<>` will disappear if there is a gap with none of them active, but `process_obs` will remain in n=1, and in an n=2 window all `get_obs<>` will remain visible.

dpmatthews: I'm finding it hard to think what would be best - nothing is perfect.
I suggest we come back to this later when we have more time & experience with the initial implementation.

hjoliver: Agreed

oliver-sanders (approved):
The best bet may well be to try the simple option first and see how people react to it.

dpmatthews (approved):

hjoliver:
Things can be merged after @dwsutherland has cast an eye over it - but he's currently engaged with more important things!

docs/proposal-n-distance-window.md
dwsutherland: If tasks with partially satisfied prerequisites exist in the pool (i.e. we don't check the DB (or w/e) for spawn conditions for each child), then we'd need to keep using suicide triggers.. Wouldn't we? i.e.
```
A:fail & B:fail => C
A | B => !C
```

hjoliver: No, the idea just above was to put those in n=1, not n=0, so (if we did that) it would be a datastore-only thing, not the task pool. Then they would presumably get pruned from the datastore once the reason for putting them there no longer applied (which is nothing to do with whether or not their prerequisites get satisfied).

dwsutherland: > datastore-only thing, not the task pool.

But the task is still in the task pool, and you'd need a despawn condition(s) to clear it from there. I see making them n=1 as a workaround to the underlying issue. Why not handle it internally and have despawn conditions as well as prerequisites? i.e. with
```
A:fail & B:fail => C
```
So C will have both `A:succeed` or `B:succeed` as despawn conditions to clear them from the pool.

Then we hit two birds with one stone (as the data-store will follow suit).. Maybe I'm missing something?

hjoliver: I see use of suicide triggers as something to be desperately avoided, except in the one(?) edge case that we possibly can't avoid for the moment - and for backward compat with Cylc 7 suites that already use them. Because it requires users to understand the difference between a task proxy in the task pool (which is purely a Cylc internal implementation thing) and the same task in the abstract graph - and all they should need to care about is the latter.

Have you read the "prerequisite housekeeping" section of the SoD proposal?  These "tasks" (prerequisites that happen to be stored in a task proxy) are not in "the task pool". They may accumulate in certain conditions and need housekeeping for scheduler memory footprint reasons, but should still be invisible to the user.  The housekeeping (which you're calling "despawn conditions" isn't so simple though:
- the conditions might not all get satisfied (they could be on alternate paths)
- your logic above uses the opposite of the spawn conditions to infer when to housekeep, but that only works for succeed and fail (which are mutually exclusive final outputs).  What about custom outputs?  (But note in the SoD proposal I do say we could potentially remove them when/if *all parents finish* (succeed or fail) because then you know the child is not going to be automatically satisfied any more.  But again, it's not guaranteed that all parents will execute at all.

hjoliver: To reiterate, these partially satisfied proxies are "in the scheduler" but not "in the task pool".  (At least that will be the case once the relevant extant Issues are addressed!)

hjoliver: Plus my "Another idea:" paragraph above, which is presumably what you're responding to, is purely about what to display in the UI in the case - Dave's example - of partially satisfied prereqs due to a family trigger - that's entirely independent of whether or not there are task proxies in the scheduler (task pool or otherwise).  If we went with that suggestion, we'd have to tell the datastore "this task in the graph has partially satisfied prerequisites so display it as n=1".

dwsutherland: > I see use of suicide triggers as something to be desperately avoided, except in the one(?) edge case that we possibly can't avoid for the moment - and for backward compat with Cylc 7 suites that already use them. Because it requires users to understand the difference between a task proxy in the task pool (which is purely a Cylc internal implementation thing) and the same task in the abstract graph - and all they should need to care about is the latter.

Yeah, I completely agree..

> They may accumulate in certain conditions and need housekeeping for scheduler memory footprint reasons

So kick the can down the road?

> the conditions might not all get satisfied (they could be on alternate paths)

Not sure I understand, the data-store will clean these alternate paths up if they no longer connect to an active task.

> What about custom outputs?

They don't have a graph opposite.. But if there's a custom graph opposite, then I get your point. . .

> we'd have to tell the datastore "this task in the graph has partially satisfied prerequisites so display it as n=1".

Feels we're "passing the buck" to the data-store, where it would take care of it properly if the pool was cleared up...

I'll have to have another think about this.. I'll be back later (dealing with kids)..

dwsutherland: > What about custom outputs?

Actually... `:finished` would be the graph opposite to any custom output, no?

hjoliver: > > They may accumulate in certain conditions and need housekeeping for scheduler memory footprint reasons
>  So kick the can down the road?

Yes, kick it out of user-space into scheduler-internals-space where it belongs.

> They don't have a graph opposite.. But if there's a custom graph opposite, then I get your point. . .

Perhaps I should have said custom output triggers: `foo:out1 => bar; foo:out2 => baz` etc.

> Feels we're "passing the buck" to the data-store ...

Well yes - the job of the scheduler is to schedule as efficiently as possible, and under SoD that's done with a very stripped-back task pool (approx. active tasks only)  The job of the datastore (including the UIS one) is to construct and serve to the UI **everything else** about the workflow that the user wants to see.

> ..., where it would take care of it properly if the pool was cleared up..

Hmm, I wonder if you're still thinking that these PST (Partially Satisfied Tasks - getting tired of typing that) are gonna be in the task pool - they're not, they will be nothing but an internal implementation detail of the scheduler (used for tracking prerequisites as they get satisfied) and as such the task pool IS cleaned up. The discussion above (to my mind) was just about - see previous point about the job of the datastore - whether or not the datastore should show extra bits of the graph beyond the strict n-window based on active tasks, in order to make it easier on users.  That would not require putting these PST back in the task pool.  

Happy to thrash this out more tomorrow if we're still not quite on the same wavelength but perhaps move it to Riot (or a VC chat).

hjoliver: > > What about custom outputs?
> 
> Actually... `:finished` would be the graph opposite to any custom output, no?

Sort of, yes - and hence this addendum to the prerequisite housekeeping section of the SoD proposal: https://cylc.github.io/cylc-admin/proposal-spawn-on-d.html#possible-stuck-prerequisite-housekeeping 

I think we are on the same wavelength about this if by "despawn conditions" above you mean "prerequisites for task proxy removal".  These would have to be stored in the child proxy like normal prerequisites and updated whenever a parent finishes (in fact I did this in an earlier SoD implementation).  However, the point I'm trying to make is that either way this will be irrelevant to the datastore and UI, because these PST objects are just scheduler internals and will not be stored in the official task pool.  If we can do the housekeeping automatically the scheduler will not have a memory leak; if we can't it might - but provisionally we decided not to bother because its difficult and it seems like all/most? cases where this would be a problem are essential suite design errors - in which case we can just alert the user to the growing number of partially satisfied prerequisites that we are keeping (in case they ever become fully satisfied) and say sort your design out!

hjoliver: > These would have to be stored in the child proxy like normal prerequisites a...

(Note my SoD section link above proposes another simpler idea too: check parent status via the DB infrequently, if and when housekeeping becomes desirable)

docs/proposal-n-distance-window.md
```diff
@@ -0,0 +1,100 @@
...
+
+PROPOSE: implement (1) initially, and later consider (2) and (3) as
+enhancements. (2) looks best but it may be tricky avoid wrongly excluding a
+child from `n = 1` in alternate path graphs?
+
+## Graph Isolates
+
+Windowing will often result in multiple disconnected graphs in the graph view
+(even without considering reflows). We will just have to live with that
+(because, see Justification above).
```
dwsutherland, line 97: These are accounted for in the n-window PR; boundary pruning is essentially `max([isolate-chain-length, n+1])`, same with end-of-branch/childless anywhere..  They will appear, then disappear after run, showing no future beyond themsel-f/ves.

dwsutherland (approved):
LGTM, just a few comments on how the current PR addresses some of the proposal.

PR #114, opened 2020-10-22
rose suite-run: proposal amendment

oliver-sanders:
Record the source directory in the "root" flow run directory as opposed to named run directory.

* This ensures all named runs are from the same installation source.
* Reinstallation knows where to install from.
* If we were to make the mandatory named run pattern optional this would be necessary to tell a named run from
  a workflow.

I've proposed using `.cylc-install/source` to record this info, though happy to change this. Perhaps we should make this info more transparent on the FS?

### DIFF
--- a/docs/proposal-rose-suite-run.md
+++ b/docs/proposal-rose-suite-run.md
@@ -102,11 +102,17 @@ The remaining points can happen any time after 1) is complete although note that
 `cylc install` - new command to install workflows.
 * By default, `cylc install` will install the workflow found in `$PWD` into
   `~/cylc-run/$(basename $PWD)/runN` (where `runN` = `run1`, `run2` ...).
+* This will create a directory in `~/cylc-run/<flow_name>` with a
+  `_cylc-install` diretory containing a `source` symlink to the source dir.
+* This will fail if the specified source directory does not match the
+  `~/cylc-run/<flow_name>/source` symlink. Otherwise it would be possible
+  for `run1` to come from a different source dir to `run2`.
 * If `--run-name` is not specified and `run1` already exists it will install
   into `run2` (and so on).
   Create a symlink `runN` pointing at the latest run.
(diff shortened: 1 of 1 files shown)


hjoliver (approved):
Agreed, seems sensible (and probably necessary).

hjoliver:
>  Perhaps we should make this info more transparent on the FS?

Is there any particular reason to hide it?  If not, I vote for transparency. It might occasionally help users to figure out where their old run directories came from.

oliver-sanders:
Happy to remove the preceding `.`, should we replace it with another prefix to distinguish it from named runs / make its intention more obvious?

* `~cylc-install`
* `-cylc-install`
* `?`

It is probably a good idea for us to have a directory at this level anyway, it may be useful later (e.g. for version control).

oliver-sanders:
Renamed `.cylc-install` to `_cylc-install` and added a note marking `_cylc_install` as being an invalid run name.

docs/proposal-rose-suite-run.md
```diff
@@ -117,8 +123,8 @@ The remaining points can happen any time after 1) is complete although note that
     Everything else will be copied.
     * If needed we could make this configurable via a `.cylcignore` file (future
       enhancement).
-  * The installation should fail if `log`, share` or `work` exist in the source
-    directory.
+  * The installation should fail if `log`, share` or `work` exist in the
+    source directory.
```
dpmatthews, line 127: I assume `_cylc-install` should now be added to this list.
(also, note the missing "`" which is presumably my fault)

oliver-sanders: It doesn't have to be in the list as the presence of a flow.cylc file would mark this as a workflow.

dpmatthews: The `_cylc-install` directory will be created in the same directory as the workflow if there is no run directory (`--no-run-name`) so we can't allow  a _cylc-install directory to be copied from the source.

dpmatthews (approved):

PR #115, opened 2020-10-27
proposal: universal id

oliver-sanders:
Adopt a universal identifier throughout all Cylc projects:

* Unblocks https://github.com/cylc/cylc-flow/issues/3592
* Solves the problem of `REG` being an optional positional argument *before* required positional arguments
  (from the rose suite-run proposal).
* Provides a more powerful and simpler CLI.
* Clarifies to users whether commands are operating on the src or run directories.

Ping @dwsutherland

### DIFF
--- a/docs/proposal-universal-id.md
+++ b/docs/proposal-universal-id.md
@@ -98,29 +98,53 @@ flow//*/*:running
 
 Reference regex:
 
-https://regex101.com/r/XOOpUW/2
+https://regex101.com/r/XOOpUW/3
 
 ```regex
-^
+# don't match an empty string
+(?=.)
+# either match a user or the start of the line
 (?:
-  ~(?P<user>)[^\/]+
-  \/?
-)?
+  (?:
+    ~
+    (?P<user>[^\/:\n]+)
+    # allow the match to end here
+    (\/|$)
+  )
+  |^
+)
 (?:
-  (?P<flow>.*?(?=(?:\/\/|$|:)))
-  (?P<flow_sel>.*?(?=(?:\/\/|$)))
-  (?:\/\/
-    (?:  
-      (?P<cycle>[^\/:]+)
-      (?::(?P<cycle_sel>[^\/]+))?
+  (?P<flow>[^\/:\n~]+)
+  (?:
+    :
+    (?P<flow_sel>[^\/:\n]+)
+  )?
+  (?:
+    //
+    (?:
+      (?P<cycle>[^\/:\n]+)
       (?:
-        \/
-        (?P<namespace>[^\/:]+)
-        (?::(?P<namespace_sel>[^\/]+))?
+        :
+        (?P<cycle_sel>[^\/:\n]+)
         (?:
-          \/
-          (?P<job>[^\/:]+)
-          (?::(?P<job_sel>[^\/]+))?
+          /
+          (?:
+            (?P<task>[^\/:\n]+)
+            (?:
+              :
+              (?P<task_sel>[^\/:\n]+)
+            )?
+            (?:
+              /
+              (?:
+                (?P<job>[^\/:\n]+)
+                (?:
+                  :
+                  (?P<job_sel>[^\/:\n]+)
+                )?
+              )?
+            )?
+          )?
         )?
       )?
     )?
--- a/docs/proposal-universal-id.md
+++ b/docs/proposal-universal-id.md
@@ -41,7 +41,6 @@ which added yet another format:
    for mutations
 5) The CLI is now implemented via GraphQL anyway meaning that the CLI arguments
    are provided as GraphQL arguments and transformed as required.
-6) We should expose users to the concept of the identifier in the UI.
 
 ## Proposed Universal Identifier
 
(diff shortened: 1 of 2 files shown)


docs/proposal-universal-id.md
```diff
@@ -0,0 +1,269 @@
...
+   ```console
+   $ cd ~/cylc-run/my_workflow
+   $ cylc stop
+   ```
+4) The UI currently doesn't make use of any form of identifier, however,
+   it may be beneficial to expose users to the global identifier e.g.
+   for mutations
+5) The CLI is now implemented via GraphQL anyway meaning that the CLI arguments
+   are provided as GraphQL arguments and transformed as required.
+6) We should expose users to the concept of the identifier in the UI.
```
hjoliver, line 44: 6 is redundant with 4?

docs/proposal-universal-id.md
```diff
@@ -0,0 +1,269 @@
...
+* 8.0bX - universal id
+  - Move to universal id in later beta release.
+  - Conversion simple in the code thanks to the conversion of the CLI to
+    GraphQL.
+  - Upgrading the tests and docs more time consuming.
+* 8.0.0 - bash autocompletion
+  - Remove all remaining `task.cycle`, `cycle/task` references.
+* 8.0.0 - bash autocompletion
+  - Not required for functionality
+  - Aim to make the 8.0.0 release to aid transfer from Rose
```
hjoliver, line 266: Duplicate sections?

hjoliver (approved):
Great, let's do it. :+1:  

Solves a bunch of problems, and doesn't create new ones that I can think of, except that we'll need a back-compat converter for Cylc 7 syntax to avoid breaking existing suites?

oliver-sanders:
Added a new "back-compat" section with an overview of the effected commands and an outline for supporting the Cylc7 syntax until Cylc9.

docs/proposal-universal-id.md
```diff
@@ -0,0 +1,324 @@
...
+  other than "user".
+  (i.e. a direct replacement for the current `cylc trigger flow *:failed`
+  syntax).
+
+```bash
+# regular form (for identifying objects)
+[~user[/]]flow//cycle/task/job
+
+# expanded form with selectors (for quick querying of objects)
+[~user[/]]flow[:state]//cycle[:state]/task[:state]/job[:state]
```
dwsutherland, line 62: > `[~user[/]]flow`

What if flow has `/` in it (path hierarchy structure) .. How do you know if `user` has been included? is it the `~`?

> `cycle[:state]`

I thought we ruled out `:` colon? Because some ISO datetime uses it

docs/proposal-universal-id.md
```diff
@@ -0,0 +1,324 @@
...
+flow//*:running
+
+# all running tasks in foo (wid="flow|*|*", state="running")
+flow//*/*:running
+```
+
+> **Note:** All syntax implemented here represents functionality currently
+  supported in the GraphQL schema.
+
+> **Note:** Currently in GraphQL `state` is a separate argument, there is no
```
dwsutherland, line 95: Currently GraphQL can do both, this was intentional.. I can't recall the exact reason, but I think it was so you could do blanket exclusions and/or inclusions of states while also picking specific named tasks (?)

docs/proposal-universal-id.md
oliver-sanders: > is it the ~?

Yep.

> I thought we ruled out : colon? Because some ISO datetime uses it

I think we should be good to use the colon because Cylc doesn't use colons for its cycle point format, and can't because colons are not safe in file paths.

docs/proposal-universal-id.md
```diff
@@ -0,0 +1,351 @@
...
+cylc stop my_workflow
+# currently: cylc stop my_workflow
+cylc stop my_workflow//123
+# currently: cylc stop my_workflow 123
+cylc stop my_workflow//123/task
+# currently: cylc stop my_workflow 123/task (or task.123)
+
+cylc trigger OPTIONS ID
+cylc trigger my_workflow//123/task
+# currently: cylc stop my_workflow 123/task (or task.123)
```
dpmatthews, line 182: Typo: should be cylc trigger not stop

docs/proposal-universal-id.md
```diff
@@ -0,0 +1,351 @@
...
+cylc trigger my_workflow//123/*:failed
+# currently: cylc stop my_workflow 123/*:failed
```
dpmatthews, line 184: Ditto

docs/proposal-universal-id.md
```diff
@@ -0,0 +1,351 @@
...
+
+* Unified universal ID which supports all current CLI syntax.
+* Makes use of the great functionality that we have already implemented!
+* Very little change from Cylc7, just add `//` between the `REG` and
+  `TASK_GLOB` arguments.
+* Exposes the full power of GraphQL to the user.
+* Enables multi-workflow control from the CLI via the UIS without complicating
+  Cylc Flow.
+* Enables multi-user control from the CLI via the UIS without Cylc Flow having
+  to know anything about authorisation.
```
dpmatthews, line 197: Not sure about this (but not important at this stage)

docs/proposal-universal-id.md
```diff
@@ -0,0 +1,351 @@
...
+```console
+# trigger cycles1, cycle2 and cycle3 in flow
+$ cylc trigger flow// //cycle1 //cycle2 //cycle3
+```
+
+The `//` serving as the marker to the bash completion to glob cycle points.
+
+## Back Compat
+
+We should be able to support the old legacy interface for the time being.
```
dpmatthews, line 301: I think I'd be happy to drop the legacy interface for most commands and only worry about those we think may have been used within scripts/workflows (e.g. broadcast).

dpmatthews (approved):
A few comments but happy to approve

dwsutherland (approved):
Looks good, will be nice to have the internal and external ID unified.

PR #118, opened 2021-01-07
cylc clean proposal

oliver-sanders:
> An outline of where we could go with the `cylc clean` command, very much open to ideas.

The current (Cylc7) approach to meddling with workflow files is a confusing trio of:

* `rose suite-clean`
* `rose_prune`
* `rose_arch`

These all have overlapping functionalities, they (ideally) *should* all:

* Locate workflow installations on different platforms (as required/configured).
* Have cycle awareness.
* Be able to perform cycle arithmetic.
* Be able to flush out files from within workflows

We wont get `cylc clean` up to `rose_arch` standards right away, nor should we, but we should be able to get some of the simpler cases out the way by early versions of Cylc8.

I would be keen for the development of a light-weight scaffold with plugins providing filetype support and archive functionality doing the heavy-lifting.

### DIFF
--- a/docs/proposal-cylc-clean.md
+++ b/docs/proposal-cylc-clean.md
@@ -1,0 +1,251 @@
+# Cylc Clean
+
+Work on the new `cylc clean` is already underway.
+
+This command is able to locate workflow installations on different platforms
+and delete them.
+
+Ideally we would want `cylc clean` to be able to do a lot more.
+
+One, simple example is re-running workflows, which requires moving/removing the
+log dir and database. As per the
+[rose suite-run migration proposal](https://cylc.github.io/cylc-admin/proposal-rose-suite-run.html#cli-changes)
+this shall be done by cylc clean.
+
+> you will be able to use cylc clean to clean out the log directory
+> (and whatever other directories you want removing)
+
+This means that we need an interface in `cylc clean` to perform this.
+
+We could give users a pre-defined option, however, the exact requirements for
+this niche usage are hard to predict.
+
+This proposal suggests developing more generic functionality in `cylc clean`
+for this and other purposes. It also highlights some viable possibilities for
+future functionality which could see it largely surpass `rose_prune` and
+`rose_arch`. Not of top importance now but worth considering for interface
+design reasons.
+
+related https://github.com/cylc/cylc-flow/issues/3887
+closes https://github.com/cylc/cylc-flow/issues/1159
+
+(For reference `rose_arch` is ~500 lines of Python and does some things we
+might not want to do in Cylc).
+
+## Examples
+
+Control over which installations of a flow are cleaned:
+
+```
+# remove a workflow installation on the scheduler host
+cylc clean myflow --local
+
+# remove a workflow installation on any remote task hosts
+cylc clean myflow --remote
+
+# remove a workflow installation on the scheduler host and
+# on any remote task hosts
+cylc clean myflow
+```
+
+Control over what is cleaned:
+
+```
+# delete a directory in a workflow installation
+cylc clean myflow --rm share/cycle/1234
+
+# delete two directores in a workflow installation
+cylc clean myflow --rm share --rm work
+cylc clean myflow --rm share:work  # equivalent
+```
+
+Basic glob support:
+
+```
+# delete all job log files from 2020
+cylc clean myflow --rm 'share/cycle/2020*'
+
+# delete everything
+cylc clean myflow --rm '*'
+cylc clean myflow  # equivalent
+```
+
+Listing rather than removing things:
+
+```
+# list stuff in share/cycle for cycles in 2020
+cylc clean --ls 'share/cycle/2020*'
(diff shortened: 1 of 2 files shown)


docs/proposal-cylc-clean.md
```diff
@@ -0,0 +1,251 @@
...
+
+# tar a directory in a workflow installation
+cylc clean myflow --tar share/cycle/1234
+```
+
+Group the `log` dir and `.service/db` together implicitly
+(use case `cylc play --re-run`):
+
+```
+# remove the log dir AND suite DB
```
dpmatthews, line 97: I think a common use case is to want to to revert your run directory back to how it was after you ran `cylc install`. I think the command to do that would `cylc clean myflow --rm log:work:share` which seems OK to me. Maybe add this as an example?

docs/proposal-cylc-clean.md
```diff
@@ -0,0 +1,251 @@
...
+```
+# remove a workflow installation on the scheduler host
+cylc clean myflow --local
+
+# remove a workflow installation on any remote task hosts
+cylc clean myflow --remote
+
+# remove a workflow installation on the scheduler host and
+# on any remote task hosts
+cylc clean myflow
```
dpmatthews, line 48: Just to note that, as discussed in cylc/cylc-flow#3887, we'll need a `--force` option to handle the case where the remote clean cannot be completed.

dpmatthews:
I think I'm happy with the detail of everything targetted for 8.0 (--local, --remote, --rm, --mv).
I haven't worried too much about the detail of the later stuff but the ideas looks good.

hjoliver (approved):
Looks great to me.

docs/proposal-cylc-clean.md
```diff
@@ -0,0 +1,251 @@
...
+# remove job logs in the range 2020-2021
+cylc clean myflow --rm log/job/<cycle> --cycle-range=2021/-P1Y
+
+# tar job logs for "foo" in the range 2020-2021
+cylc clean myflow --tar log/job/<cycle>/foo --cycle-range=2021/-P1Y
+
+# tar job logs for "foo" for the year before $CYLC_TASK_CYCLE_POINT inclusive
+cylc clean myflow --tar log/job/<cycle>/foo --cycle-range=<cycle>/-P1Y
+
+# delete the work dor for the task being run
```
hjoliver, line 125:
```suggestion
# delete the work dir for the task being run
```

docs/proposal-cylc-clean.md
```diff
@@ -0,0 +1,251 @@
...
+            """
+
+        # we could add a main-loop event which gets fired whenever a cycle
+        # "closes" which would be good for housekeep purposes
+
+        # can always use the same options configured here on the command line
+```
+
+```python
+# to illistrate how trivial this plugin would be ...
```
hjoliver, line 224:
```suggestion
# to illustrate how trivial this plugin would be ...
```

hjoliver (approved):
Just to play devil's advocate: maybe this is somewhat overkill for housekeeping of workflow files - is there really much need for MIME type-specific plugins, for instance, when 99%  (?) of use would be just to delete or compress and/or archive files by location, file extension, and cycle point. (Is the file type content really relevant here?)

oliver-sanders:
MIME is overkill, really just suggesting that if we add "tar" support we do so via an interface which allows support for other types (or use on if it exists). MIME is just a cheap way of getting file-type association.

oliver-sanders:
Just considering one final related use-case which is the transfer of files between platforms.

* Internally (e.g. remote_init)
* By users (e.g. to move data where it is needed)

Obviously not a priority ATM but worth ensuring the interface is compatible with this in future.

The utilities outlined here provide most of the framework for this, in particular the ability to find what platform a task has run on and to pick a host on the same filesystem where the data can be extracted. This is very hard for users to do, especially in a portable way.

Something like this:

```bash
# copy "somefile" from one platforma to platformb
$ cylc clean myflow \
  --scp share/cycle/1234/somefile platforma \
  --platform platformb

# copy "somefile" from the platform where 1234/foo ran (most recently) to the scheduler host
$ cylc clean myflow//1234/foo \
    --scp share/cycle/<cycle>/somefile \
    --platform=<scheduler>

# internal job log retrieval would look like this
$ cylc clean myflow//1234/mytask/* \
    --rsync log/job/<cycle>/<task>/<job>/ <scheduler>

# remote-init would look like this
$ cylc clean myflow \
    --scp=./ myplatform \
    --exclude='*' \
    --include=...
```

Would need a couple of subtle change to the above to fix. Also perhaps better to split into `cylc clean` and `cylc sync` (as two interfaces to the same functionality) for clarity.

hjoliver:
We should get this merged, so that the proposal document is can be found easily.

dpmatthews (approved):
Let's get this merged.
Detail of initial changes looks fine.
Future stuff can be debated further later.

PR #121, opened 2021-02-18
Update re-install proposal.

wxtim:
I want to check that I'm right in expecting this to work in the way set out below before I progress it.

### DIFF
--- a/docs/proposal-cylc-rose-installing-rose-configs.md
+++ b/docs/proposal-cylc-rose-installing-rose-configs.md
@@ -207,7 +207,7 @@ $ ROSE_SUITE_OPT_CONF_KEYS=b cylc install -O c
 
 ```
 # ~/cylc-run/FLOW/rose-suite.conf
-opts=a b c
+opts=a (cylc install)
 ```
 
 ```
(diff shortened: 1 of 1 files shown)


oliver-sanders:
With rose the `opts` setting is *only* loaded in the root `rose-<type>.conf` file, so whatever you define in `opts` in an optional configuration does nothing. This is why we ignore it (`!opts`) in the optional configs to make it clear that this is non-functional.

Here's the example from the proposal (example4) with the `cylc-install` config (example3) added:

```console
$ cat ~/roses/flow/rose-suite.conf
opts=a
$ ROSE_SUITE_OPT_CONF_KEYS=b cylc install -O c
$ cat ~/cylc-run/FLOW/rose-suite.conf
opts=a b c (cylc-install)
$ cat ~/cylc-run/FLOW/opt/rose-suite-cylc-install.conf
!opts=b c
$ cylc reinstall -O d
# ~/cylc-run/FLOW/opt/rose-suite-cylc-install.conf
!opts=b c d (cylc-install)
$ cat ~/roses/FLOW/rose-suite.conf
# with each reinstall the cylc-install|opts are appended to the end
# of the rose-suite.conf|opts setting
opts=a b c d
```

PR #125, opened 2021-04-26
Proposal cylc authorisation

datamel:
Proposal for UI-Server authorisation. 
Proof of concept PR : https://github.com/cylc/cylc-uiserver/pull/204


datamel:
Converting this to a draft as the config section in the proposal requires additional thought.

hjoliver (commented):
Looks really good so far @datamel, nice and clear.  A few things to discuss at the next meeting, if not resolved by then.

hjoliver:
I took another pass through this, still looks great :+1: 

The only thing I'm wondering about is the current read/write/execute model, where execute permission (scheduler start, stop) implies write permission (task mutations).  Arguably that should be the other way around, or perhaps separate with no inheritance of one from the other. If you can start my workflow it will run exactly what I configured it to run.  But if you can execute task mutations on my running workflow you can change what actually runs (potentially drastically).

dwsutherland:
Yeah, need something like `read`/`play`/`edit`

Where `play` would be stop/play/trigger ...etc, and `write` would be file-manipulation(viewing?)/trigger-edit/broadcast

> UI Server back-end code should prevent users manipulating client-side code to bypass authorisation. For example, sending a GraphQL query for a different operation to that stated in the operation name parameter.

As mentioned elsewhere, we can achieve fine grained authorisation via middleware. For example:
https://docs.graphene-python.org/en/latest/execution/middleware/#example
where can write any function into this, and have all information needed here (even the cached authorisation config if so desired).. And it should be the first middleware that is executed (the last thing is always the resolver fetching/command-run).

The reason we should use Graphene's inbuilt middleware, is because request object methods (like `get_graphql_params` in your PoC) only give you the operation i.e. `query`, `mutation`,  `subscription`. And you'd need to parse the graphql doc from there, because you can run something like this:
```
mutation {
  pause (workflows: ["fox/run1"]) {
    result
  }
  kill (workflows: ["fox/run1"], tasks: ["*"]) {
    result
  }
}
```
At the middle where level you could allow `pause` and do nothing (or return unauthorised) on `kill` (or w/e combination).

> I've tried a number of methods for fetching group membership for the authenticated browser users i.e. return the same groups as I see using groups on the Linux command line, which should be both local groups and remote via SSSD/LDAP.

I don't think we need to tether the grouping of our UIS authorisation to the file system groupings (as the HPC/System admins may have their own grouping ideas), but allow for users to decide what other system users can do with their suites. At NIWA, I don't have write permission or are part of the same file system grouping as our operational role accounts (but have sudo access to assume their role). Obviously the config should be user write only (0644), and this should be checked when loading it in.

It will essentially allow you to give full permission of your user and it's capabilities (as with `edit`/`execute` permission, file/job edit and broadcast can be made to do anything you like), so to please HPC/systems/security folk, you'd need to have a max giveaway configurable at the site level (defaulted to `read`).

> Not directly related to this proposal, although perhaps worth bearing in mind is access on the command line.

Once the CLI can go through the proxy => hub => UIS, we'll be covered 👍 

Nice work!

datamel:
Thanks so much for the feedback @dwsutherland. I had investigated middleware and moved away from it initially, due to other things needing authorisation (not just graphql queries/mutations). Thanks for your example, I shall have another look.
The POC is a little outdated (as the proposal has evolved), I will look at updating it and this proposal.

docs/proposal-multi-user-approach.md
```diff
@@ -36,156 +36,174 @@ UI Server back-end code should prevent users manipulating client-side code to by
...
+        },                      # server_owner1, default access for user2 falls back to
+                                # standard READ only (if server_owner_2/user2 are
+                                # included in other auth config e.g. the top example),
+                                # or none if not in any other auth config sections.
+
+        "group:<groupA>": {     # group denoted with a `group:`
+            "default": "ADMIN"  # groupA has default ADMIN access to server_owner_2's
+        },                      # workflows
+    },
+    "group:<grp_of_svr_owners>":{         # Group of users who own UI Servers
```
datamel, line 93: I am not sure of the business need for this, although included it for discussion. Do we want to support making authorisation by a group of UI Server owners possible? Perhaps this is unnecessary complexity?

hjoliver: Struggling to think of the case here too. Is it something like this?: "oper" users can authorize "test" users to access any oper server?

datamel: Yes, that is what I had envisioned, although unsure if that would be a useful feature. I guess it depends if that kind of bulk permission for authorisation/default settings has a use case.

dpmatthews: I think it's a reasonable feature to have and hopefully not too difficult since we need to support groups for specifying the users. One use case: all members of the group "my-team" grant READ access to other members of "my-team".

docs/proposal-multi-user-approach.md
```diff
@@ -13,11 +12,276 @@ via a UI Server running as the suite owner for UI purposes meaning that, from
...
+    "group:<group1>": ["ADMIN"],                # All access to workflows granted
+    "<user2>": ["ADMIN", "!trigger", "!edit"],  # All Admin, not trigger, or edit
+    "<user3>": ["!ADMIN"]                       # No privileges (incl READ)
+}
+```
+
+### Example Site Configuration
+
+```python
+c.UIServer.authorisation = {
```
dpmatthews, line 63: I think you'll need to use a different name to distinguish this from the user config.

docs/proposal-multi-user-approach.md
```diff
@@ -13,11 +12,276 @@ via a UI Server running as the suite owner for UI purposes meaning that, from
...
+
+### Site vs. User Config Precedence, Defaults and Limits
+
+At CylcCon2020 is was agreed that a:
+
+> user can ramp up authorization levels as far as the site allows
+
+Site config `limit` will determine if a user has the privileges to give away access to their workflows. It will  it will also set a `default` access level. Users cannot raise access levels in their UI Server config for a given user or group, higher than those set in site config `limit`.
+
+If the site config does not set access for a given user or group then a UI Server will **not** be limited by site config.
```
dpmatthews, line 112: This seems at odds with the deny by default principle mentioned above. I think there should be no multi-user access possible unless permitted in the site config.

datamel: Fair enough point, I will change to if a user is in the user config but not in the site config, we assume a user can not grant permission and deny?

hjoliver: That's somewhat unfortunate, but I think we do need to do it.

datamel:
So far, we have the following options.
It would be good to get some final thoughts from you all. Perhaps I need to extend my PoC for all 3?

### Option A: Decorating handler actions
Adapt my current authorization PoC so that it parses the query body to get all operations in the graphql.  I had originally identified this as something we would need to do (was aware that we could not rely on operation name alone): 

    # TODO Should also check the query e.g. it starts with 'mutation pause'
    # For security in case an operation name is sent different to the query
    # Hence fetching below

I still feel strongly that the responsibility of authentication and authorization should sit with the handler layer and prevent unauthorised access as early as possible. We expose more code, the further down we check authorization, which is not great if we’re also being a bit hacky with said code to make it do auth elsewhere.

### Option B:  Middleware
As @dwsutherland rightly suggests, we could place the auth in graphene middleware. Although we do not have immediate access to the authenticated user at this point.
So, to do this we’d push additional properties onto the middleware object in the handlers. I have tested this (obviously just hard-coded to the second element but we can find it via object type).
For the subscription handler via the subscription server field e.g. 
```def get(self, *args, **kwargs):
           self.subscription_server.middleware[1].current_user = self.current_user
           self.subscription_server.middleware[1].user_auth_config = self.user_auth_config
```
For the graphql handler via the middleware list:
```self.middleware[1].user_auth_config = self.user_auth_config
    self.middleware[1].current_user = self.current_user
```
This worked fine, although we would _still_ need to parse the graphql in the middleware to determine the operations.
This would only cover authorisation for graphql and not the static files, the main UI handler or the user profiler handler. It would also mean having 2 independent authorisation solutions, both needing maintenance etc.

### Option C: Decorating the resolver functions
As per @oliver-sanders thoughts that we should avoid additional parsing of the query, which we need to do to discover operations in a query.
My instincts when I first picked this work up was to work with the resolvers themselves but I was told that only the UI server code should be touched. So I have not had a chance to prove this.

I’m happy to give it a go for another PoC but there’s a few points for discussion first:
•	Do we really want to pass authentication that far down from the handler? Feels wrong to raise a web.HTTPError(403) from cylc flow code, which shouldn’t have ‘knowledge’ of the website side of things.
•	Might be hacky getting user details that far (not proved it yet).
•	Wouldn’t we be doing additional processing over every resolver being hit (until an unauthorized operation is reached)? - surely this would tip the argument in favour of option A which, granted, has drawback of additional parsing of the query but all three options have some additional processing debt.
•	As Option B: This would only cover authorization for graphql and not the static files, the main UI handler or the user profiler handler. It would also mean having 2 independent authorization solutions, both needing maintenance etc.


Also worth noting, I have looked into whether we can parse the request at handler and then continue passing this through, ready parsed, but the graphene layer parses the request regardless.
https://github.com/graphql-python/graphene-tornado/blob/2800ff581fd2edd665d2cf996e1601e7e892cb53/graphene_tornado/tornado_graphql_handler.py#L171.
I could put a pr into graphene to skip when it receives an already parsed request but I suspect it would not be accepted as that is major code change which could have a knock on effect for other users. So, I feel if we go with this option we are stuck parsing twice.


On a related point, if a user requests multiple mutations in one request, and they are authorised for e.g. 2 out of 3, do we want…
a)	Two to be allowed and the third to be blocked and logged
b)	The entire request to be blocked and logged, i.e. all 3 mutations denied?
I'm thinking most likely a) but thought I'd double check.

oliver-sanders:
I can see the temptation to parse the GraphQL document up front but I'll chuck in a few comments to show how this could be vulnerable.

This approach depends a lot on our ability to replicate the internal workings of Graphene as closely as possible and to keep up with their internal implementation. If we lapse the approach might not be quite as iron-clad as we would like it to be.

In parsing the document upfront we are trying to guess what endpoints that the request would hit rather than waiting for the server to tell us what endpoints are being hit. If we guess wrong then we are in trouble, here are some conditions that could break our guesses:

* We parse the document differently to the Graphene internals.
* We don't interpret the results correctly.
* The attacker finds a way to disguise a mutation or change its identifier in the request.
* The Graphene internal logic changes in a future version (we don't have control over what version the user installs).
* The GraphQL language evolves providing new representations of mutations.
* A bug in graphene causes something to be called which wasn't supposed to be.

Then, without any protections further down the chain the entire authorisation layer could be sidestepped.

In a conventional HTTP(s) server (like the Cylc 7 suite server) the server routes each request based on the URL, any auth is performed at the endpoint. In a GraphQL server the routing is based on the GraphQL document, the endpoints are the resolvers.

I found an authorisation framework for GraphQL (2.8k⭐️) called [graphql-shield](https://github.com/maticzav/graphql-shield), sadly it's Javascript (like all the GraphQL goodies) so we can't use it, however, worth noting they went down the middleware approach.

hjoliver:
I agree with @oliver-sanders on this.

dwsutherland:
I agree with @oliver-sanders too, we should not be parsing the GraphQL document/object ourselves.. For all the stated reasons.

However, we really need to be sure of what information we have and what we can do with GraphQL already, before ruling things out.. So just some clarifications on Option B:

> Although we do not have immediate access to the authenticated user at this point.

Can't we just pass this in via info/context? You simply set fields on info.context.
It may even be simpler than that, because the whole request object is also available via `info.request` (or the like).

> This worked fine, although we would still need to parse the graphql in the middleware to determine the operations.

No (sorry to be blunt). This information is all available (`info.operation.name.value`), no parsing.
For example:
```
class IgnoreFieldMiddleware:
    """Set to null/None type undesired field values for stripping."""

    # Sometimes `next` is a Partial(coroutine) or Promise,
    # making inspection for know how to resolve it difficult.
    ASYNC_OPS = {'query', 'mutation'}

    def __init__(self):
        self.args_tree = {}
        self.tree_paths = set()
        self.field_sets = {}

    def resolve(self, next, root, info, **args):
        """Middleware resolver; handles field according to operation."""
        # GraphiQL introspection is 'query' but not async
        if getattr(info.operation.name, 'value', None) == 'IntrospectionQuery':
            return next(root, info, **args)

        if info.operation.operation in STRIP_OPS:
            path_string = f'{info.path}'
.
.
.
```
There's all the information we need here.  i.e. `info.field_name`, `info.path`... etc


I think we need two levels:
- Top level: Handlers .. etc.
- GraphQL: Middleware

oliver-sanders (commented):
All looks good to me.

Just one little curveball that you may have already spotted, and which we were aware before we started on this. Commenting it here for the record and also as a reminder that we should add a note of some form to the documentation when we write it.

**These site-configured limits can be hacked and overwritten by users.**

There are at least two Cylc mechanisms by which the limits could be hacked from user-space:
* Setting `CYLC_SITE_CONF_PATH`, writing your own site config and injecting it:
  * In the spawner args (user config).
  * In shell profile files (unless Cylc is deployed in a venv of some form with a wrapper script).
  * On the cli (i.e. starting the UIS manually).
* Monkeypatching the UIS code from the config file.
  * Because the config file is in Python and you can change anything in Python.
  * We could switch to a JSON config file, but, Jupyter server supports Python config files and we can't turn that support off.

These sound like security bugs, but they are not. Preventing these mechanisms from being exploited would be pseudo security. There are plenty of non Cylc-specific mechanisms too:

* Monkeypatch the UIS code from the `PYTHONSTARTUP` file.
  * This applies to ALL Python programs!
* Wrap the Cylc UIS code to create your own uis command...
  * ...and get the spawner to run this in the user config.
  * ...and patch it by setting PYTHONPATH.
* I'm sure many OS level approaches.

On the surface this might sound bad, but it isn't because:

* The UIS runs in user space, users have control over this, there is nothing Cylc can do to stop them.
* Users can always give away full access to their accounts by sharing credentials (short of biometric security even 2FA can't prevent this).
* The only way to truly achieve this would involve running the entire UIS under a site-controlled privileged account.

So in the same way that "not giving away your user credentials" or "not using infrastructure for malicious purposes" are enforced as human policies "not hacking Cylc to fiddle the site-defined limits" is a human factor too.

docs/proposal-multi-user-approach.md
```diff
@@ -13,11 +12,344 @@ via a UI Server running as the suite owner for UI purposes meaning that, from
...
+Unset defaults for both the `limit` and `default` will need consideration, I suggest that, in keeping with the deny by default principle:
+
+* if a limit is not set but a default is, then the default level is used as the limit.
+* if a default is not set but a limit is, then the default should be no access.
+
+If a user does not appear in a ui-server owners config, then the default access from site config will apply. Otherwise, defaults will not impact upon access.
+
+We will need to consider the desired behaviour if a user appears twice with different defaults and limits set, this is probably most likely to occur when a user appears in either multiple groups or in a group and as a user.
+
+Proposal: all permissions are additive, if user appears elsewhere in config, the permission level should be taken as the greatest possible. Note, default in site config does not contribute to permissions, unless user does not appear in ui server owner's config.
```
hjoliver, line 129: That's sensible. If the higher level of permission is granted in the config, then no one can rightfully complain about us granting it (even if lower levels are explicitly granted as well).

docs/proposal-multi-user-approach.md
```diff
@@ -13,11 +12,344 @@ via a UI Server running as the suite owner for UI purposes meaning that, from
...
+
+## Possible Access Assignment of Mutations and Queries
+
+### Queries
+
+Proposal: All queries can be executed if the user has `Read` access.
+
+### Mutations
+
+If running with the read/admin configuration, initial assignments will be needed, for the case when users set e.g. when users assign: `READ`, `ADMIN`
```
hjoliver, line 185: ?
```suggestion
If running with the read/admin configuration, initial assignments will be needed, for the case when users set e.g.: `READ`, `ADMIN`
```

docs/proposal-multi-user-approach.md
```diff
@@ -13,11 +12,344 @@ via a UI Server running as the suite owner for UI purposes meaning that, from
...
+
+Some of the below are not currently available to users but including them here for consideration. This is also a fairly substantial list, which makes the case for read and admin pre-set access groups which would be easier for users and sites to configure, rather than defining a long list of mutations per user/group.
+
+As a springboard for discussion, defaults could be assigned as follows:
```
hjoliver, line 189: Looks to me like you've nailed it, as-is.

docs/proposal-multi-user-approach.md
```diff
@@ -13,11 +12,344 @@ via a UI Server running as the suite owner for UI purposes meaning that, from
...
+In code, `READ`, `ADMIN` and `ALL` will expand to the assigned operations.
+
+Any mutation without a specified access assignment will be denied by default.
+As future features are added, they will also need to be categorised.
+
+The associated arguments for the mutations may need consideration.
+
+### Security Check for Permission Level on Config Files
+
+We should set a recommended write permissions level for the user config file. It could pose a security risk to have users config files writable by others. Perhaps we could have a warning in the log if the file permissions are not strict enough.
```
hjoliver, line 253: We should probably log a warning that the file is not sufficiently secure (writable by others) and fall back to locked down permissions (i.e., owner only).

docs/proposal-multi-user-approach.md
```diff
@@ -13,11 +12,344 @@ via a UI Server running as the suite owner for UI purposes meaning that, from
...
+            }
+        },
+
+```
+
+## Open Config Questions
+
+* Config design model: CylcCon read/write/execute vs READ/ADMIN.
+
+* Access group defaults (suggested in table above) need confirmation/agreement of mapping operations/mutations to correct groups. This is something, that although technically can be changed with ease, would potentially be very confusing for users if changed at a later date.
```
hjoliver, line 274: Let's see if we can agree on these two in today's project meeting.

hjoliver: BTW I agree with @oliver-sanders that ADMIN should be renamed to CONTROL.

hjoliver:
Excellent diagrams :tada:

docs/proposal-multi-user-approach.md
```diff
@@ -13,11 +12,344 @@ via a UI Server running as the suite owner for UI purposes meaning that, from
...
+## Not supported in this proposal
+
+* Other possible configured attributes such as time-based authorisation. e.g. limit access to UI Server to User_A for 30 minutes.
+
+Whilst this would be useful to help users debug their workflows, screen sharing through Teams, for example, can currently provide access to a users UI Server and workflows. The control can be handed over to another user but the owner/user can remove access control at any time.
+
+* Workflow level granularity
+
+Initially this authorisation work will be implemented on an all workflow basis, i.e. grant User_A access to all my workflows.
+Future work could implement authorisation configuration on a per-workflow basis.
```
hjoliver, line 336: I wonder if we should provide a workflow placeholder in the hierarchy of authorization config items from the start though, even if it initially can't be modified (from `*` or whatever), to avoid breaking changes when we introduce this capability?  (Did we think about how to incorporate workflow specificity already - I don't recall?)

hjoliver (commented):
Nice proposal :+1: Just a handful of minor comments and suggestions.

docs/proposal-multi-user-approach.md
```diff
@@ -13,11 +12,344 @@ via a UI Server running as the suite owner for UI purposes meaning that, from
...
+        },                      # server_owner2, default access for user2 falls back to
+                                # standard READ only (if server_owner_2/user2 are
+                                # included in other auth config e.g. the top example),
+                                # or none if not in any other auth config sections.
+
+        "group:<groupA>": {                # group denoted with a `group:`
+            "default": ["READ", "ADMIN"]   # groupA has default READ, ADMIN access to server_owner_2's
+        },                                 # workflows
+    },
+    "group:<grp_of_svr_owners>":{          # Group of users who own UI Servers
```
hjoliver, line 96: How are these groups to be defined?

datamel: These are ui-server owner members of system groups. The groups will be returned by an `os.getgrouplist` call.

docs/proposal-multi-user-approach.md
datamel: No discussions as yet on the design of this but it could take the form...
```python
c.UIServer.authorisation = {
    "User1": {'flowA': ["READ", "ADMIN"],
               'flowB': ["READ", "play", "stop"],
               '*': ['READ']
    }
}
```
I guess implementing this with only the `*` as you suggest would leave the potential for adding it later.

oliver-sanders: > Did we think about how to incorporate workflow specificity already - I don't recall?

Yes we did, however, it could only make sense for read permissions since control of one workflow could open up control of all workflows.

hjoliver: In principle yes, but if you're granting a colleague control of one of your workflows it is probably reasonable to trust that they won't use trigger edit (say) to alter tasks in it in order to deliberately mess with your other workflows!  So I think we should have that capability, and just document that the potential for danger there.

datamel: Just noting here that, as discussed in the VC today, workflow configuration will not be implemented (proposal also updated for future reference).

docs/proposal-multi-user-approach.md
datamel: Proposal updated.

docs/proposal-multi-user-approach.md
```diff
@@ -44,71 +44,68 @@ The config discussed at [CylcCon2020](https://github.com/cylc/cylc-uiserver/issu
 
 ### READ and ADMIN
 
-For the purposes of clarity for users, and since we have operation (mutation) level granularity, the simpler method of having two groups, `READ` and `ADMIN`, rather than three. Access groups should be entered in caps to avoid confusing e.g. READ with read mutation.
+For the purposes of clarity for users, and since we have operation (mutation) level granularity, the simpler method of having two groups, `READ` and `ADMIN`, rather than three. These two access groups can be combined for ease of configuration with `ALL`. Access groups should be entered in caps to avoid confusing e.g. READ with read mutation.
```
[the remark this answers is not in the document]
oliver-sanders: Thinking back to previous discussions, I think we wanted to put "broadcast" into an "edit" category?

(note tha Cylc8 trigger-edit functionality will be a combination of broadcast+trigger)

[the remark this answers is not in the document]
datamel: Are you thinking to have `READ`, `CONTROL` and `EDIT`?
Where `EDIT` has both `broadcast` and `trigger` in it?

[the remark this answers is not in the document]
oliver-sanders: `EDIT` would just have `broadcast` in it (because broadcast allows us to change what the workflow runs), `trigger` is a regular `CONTROL` cause all it does is run what is defined in the `flow.cylc` file. `trigger-edit` is a Cylc 7 thing that we are replacing with a combination of broadcast + trigger.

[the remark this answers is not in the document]
dpmatthews: Not sure it's worth having `EDIT` if it only has one thing (can always add it later).
If we want `CONTROL` to only allow users to interact with existing workflows then I think we also ought to exclude `install` and `re-install` (also `edit` and `terminal access` although I'm not sure we'll be able to provide multi-user support for these anyway). `reload` and `play` also have the potential to change the intended workflow behaviour but I think we have to accept that.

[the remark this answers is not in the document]
oliver-sanders: I think it's worth it if it makes the corruptive nature of broadcast more apparent to sites which have to explicitly opt in to allowing users to give away that permission so can't be caught unaware. Doesn't matter that it's the only mutation in this group, I'm sure we can come up with more later!

Uninstalled workflows are still "existing workflows" so I'm not sure that there's a good reason to move them out of "control". You can only provide install options that are supported by the workflow anyway. Same for reload and play.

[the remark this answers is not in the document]
oliver-sanders: (sidenote: from a security perspective broadcast and full terminal access are no different, they both allow unchecked arbitrary code execution so if we are going to allow one there shouldn't be a barrier to allowing the other).

oliver-sanders (approved):
:+1:

docs/proposal-multi-user-approach.md
```diff
@@ -13,11 +12,350 @@ via a UI Server running as the suite owner for UI purposes meaning that, from
...
+
+### Config Design
+
+The config discussed at [CylcCon2020](https://github.com/cylc/cylc-uiserver/issues/10#issuecomment-773752542) suggests admin levels of `read`, `write` and `execute`. Reducing the number of access groups is another suggestion, especially since write and execute are not very meaningful to users (in terms of Cylc) and to ensure users do not accidentally give away more privileges than intended, a simpler system, as outlined below, may be preferential. Proposal: `READ` and `CONTROL` permission groups.
+
+### READ and CONTROL
+
+For the purposes of clarity for users, and since we have operation (mutation) level granularity, the simpler method of having two groups, `READ` and `CONTROL`, rather than three. These two access groups can be combined for ease of configuration with `ALL`. Access groups should be entered in caps to avoid confusing e.g. READ with read mutation.
```
dpmatthews, line 47: Suggest replacing second sentence with:
A further group, `ALL`, can be used to provide full access (`READ` + `CONTROL` + high risk operations).

docs/proposal-multi-user-approach.md
```diff
@@ -13,11 +12,350 @@ via a UI Server running as the suite owner for UI purposes meaning that, from
...
+
+```python
+c.UIServer.authorisation = {
+    "<user1>": ["READ", "pause", "trigger", "message"], # Specified interactions granted to user1
+    "group:<group1>": ["ALL"],                          # Both READ and CONTROL access to workflows
+                                                        # granted to users in group1
+    "<user2>": ["READ", "CONTROL", "!trigger", "!edit"],  # All READ and CONTROL, except trigger and edit
+                                                        # granted to user2
+    "<user3>": ["READ"]                                 # READ only access
+    "<user4>": ["!READ", "!CONTROL"]                      # user4 has no access to workflows (same as !ALL)
```
dpmatthews, line 59: Change to "!ALL" (["!READ", "!CONTROL"] no longer equivalent)

docs/proposal-multi-user-approach.md
```diff
@@ -13,11 +12,350 @@ via a UI Server running as the suite owner for UI purposes meaning that, from
...
+
+```python
+c.UIServer.site_authorisation = {
+    "<*>": {                    # For all ui-server owners,
+        "<*>": {                # Any authenticated user
+            "default": "READ",  # Will have default read-only access
+        },
+        "<user1>": {              # user1
+            "default": [
+                "!READ", "!CONTROL" # No privileges for all ui-servers
```
dpmatthews, line 73: Change to "!ALL" (["!READ", "!CONTROL"] no longer equivalent)

docs/proposal-multi-user-approach.md
```diff
@@ -13,11 +12,350 @@ via a UI Server running as the suite owner for UI purposes meaning that, from
...
+            ],                    # owners. Note, could also have configured as !ALL
+        },                        # No limit set, so all ui-server owners
+    },                            # limit is also "!READ", "!CONTROL" for user1
+
+    "<server_owner_1>": {                 # For specific UI Server owner,
+        "<*>": {                          # Any authenticated user
+            "default": "READ",            # Will have default read-only access
+            "limit": ["READ", "CONTROL"]    # server_owner_1 is able to give away
+        },                                # all privileges.
```
dpmatthews, line 82: READ and CONTROL

docs/proposal-multi-user-approach.md
```diff
@@ -13,11 +12,350 @@ via a UI Server running as the suite owner for UI purposes meaning that, from
...
+    },
+    "<server_owner_2>": {       # For specific UI Server owner,
+        "<user2>": {            # Specific user2
+           "limit": "ALL"       # Can only be granted a maximum of ALL (READ and CONTROL) by
```
dpmatthews, line 86: Not READ and CONTROL

dpmatthews (commented):
Several minor changes needed now ALL is not the same as READ + CONTROL

hjoliver (approved):

hjoliver:
:tada:

PR #128, opened 2021-07-16
optional trigger proposal (rejected)

dpmatthews:
Propose way forward for cylc/cylc-flow#4256

### DIFF
--- a/docs/proposal-partially-satisfied-prerequisites.md
+++ b/docs/proposal-partially-satisfied-prerequisites.md
@@ -1,0 +1,240 @@
+# Handling of partially satisfied prerequisites
+
+See https://github.com/cylc/cylc-flow/issues/4256
+
+## Partially satisfied prerequisites
+
+Should tasks with partially satisfied prerequisites cause a stall if the scheduler has nothing else to run?
+Currently this is not the case with Cylc 8.
+
+Consider the following workflow:
+
+```ini
+[scheduling]
+    [[graph]]
+        R1 = """
+             foo => bar => qux
+             foo:fail => baz => qux
+             """
+```
+
+Currently this will just run `foo` and `bar` and then shutdown.
+The user intended `qux` to run but has made a mistake in the graph.
+It's too easy for a user to miss this (especially if this is part of a complex workflow or if it's something that only happens in some failure scenario).
+For example:
+
+```ini
+[scheduling]
+    cycling mode = integer
+    initial cycle point = 1
+    final cycle point = 5
+    [[graph]]
+        P1 = """
+             model[-P1] => model => archive
+             archive[-P1] => archive
+             archive:fail => recover
+             """
+[runtime]
+    [[model,archive,recover]]
+        script = sleep 2
+    [[archive]]
+        pre-script = [[ $CYLC_TASK_CYCLE_POINT != 2 ]] || exit 1
+```
+
+A simple workflow where a recovery task has been added to deal with the rare case where the archiving fails.
+However, a mistake in the graph means the archiving will be skipped in any cycle after a failure.
+This kind of error is just too easy to make in complex suites - we need to stall in this situation.
+
+Similar situations could occur operationally when they have to intervene due to problems.
+Some tasks might get skipped in one cycle which could then lead to partially satisfied prerequisites in the next cycle - these require intervention.
+
+Reload is another potential cause for issues.
+Graphing changes to a running workflow could easily result in inconsistencies which cause partially satisfied prerequisites in the initial cycles.
+Imagine making an operational change to the graph which then resulted in part of the graph being skipped - confidence in Cylc would quickly be destroyed.
+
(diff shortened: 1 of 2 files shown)


docs/proposal-partially-satisfied-prerequisites.md
```diff
@@ -0,0 +1,347 @@
+# Handling of partially satisfied prerequisites
+
+See https://github.com/cylc/cylc-flow/issues/4256
+
+## New trigger syntax
+
+A new trigger syntax is proposed which (it is hoped) will resolve many issues:
+
+* `a:x ?=> b` is an "optional trigger": `b` is optionally triggered by `a`.
```
wxtim, line 9: I think that `a:<state> ?=> b` would be more intuitive to the reader.

oliver-sanders: This is referencing the earlier example which used `a:x => b`.

hjoliver:
(I'd suggest digesting #129 now, before getting too mired in the details of this PR)  (Although @oliver-sanders xtrigger ruminations above probably apply equally to both proposals).

hjoliver:
**This proposal rejected in favour of #129**

(Can close it after transferring the xtrigger comment above elsewhere)

docs/proposal-partially-satisfied-prerequisites.md
```diff
@@ -0,0 +1,347 @@
...
+        a | r1 => b ?=> r2
+        a:fail ?=> r1 => r2
+```
+
+In this case, recovering from the failure of `a` involves running `r1` before running `b` and then running `r2` once `b1` has succeeded.
+If `a` succeeds then `r2` will have partially satisified dependencies but this is not a problem since it is optionally triggered.
+Note that this is an example of where a task is optionally triggered by one task but required by another.
+This is not a problem.
+`r2` only becomes required if `r1` is run.
+
```
hjoliver, line 125: Response to this example for the record. 

This is actually a pathological graph for Cylc 7 and 8: it requires the same tasks to have different dependencies on alternate branches, but our graph parsing assumes all instances of a given task name in the graph refer to the same task, and we give the task definition all of the harvested dependencies.

We intend to fix this in Cylc 9 with dynamic subgraphs:
```
a:succeed => {r1 => b => r2}
a:fail => {r1 => r2}
```
In the meantime, for Cylc 7 and 8 we have to use different logical tasks to represent the same job content on different branches.

Also, the proposed trigger syntax is hard to interpret here. Counter-intuitively the optional trigger `b ?=> r2` does not correspond to an optional output of `b`. Instead it means *"If b runs we will spawn r2, but r2 can only run if it also gets spawned by other non-optional triggers"*.  Worse, the other non-optional trigger is an artificial dependency that won't
come naturally to users: `r1 => r2` (the natural chain is `r1 => b => r2`).

That strikes me as a convoluted bodge that works for the wrong reasons. We'd be using optional and artificial triggers to get around the nasty fact that `b` has accumulated all the triggers from two different mutually exclusive branches.
So I wouldn't be surprised if we could dream up cases that fail the bodge.

Also, we'd end up with a partially satisfied waiting `r2` every time the usual success branch runs, and I'm not entirely convinced we could prune that immediately if the unsatisfied prerequisite is marked "optional". What if the
two branches aren't mutually exclusive? The optional trigger might still get satisfied later on...

hjoliver:
Difference from Optional Outputs #129 

Consider `a:x => b` where `a` completes without generating `a:x`.

- With expected triggers: we remove `a` as completed but spawn `b` into the n=0
  pool anyway as an *unsatisfied waiting task*, and flag that as a problem
  - the user has to figure out why `b` is waiting (and the solution likely
    involves fixing and rerunning the upstream task `a`)

- With expected outputs: we do not spawn `b` but we retain `a` in the n=0 pool
  as an *incomplete task*, and flag that as a problem
  - the user can see `a` did not generate an expected output (and the solution
    likely involves messing with the flagged task `a`)
  - we do not spawn waiting children that (depending on the resolution of the
    problem) might not be needed

docs/proposal-partially-satisfied-prerequisites.md
```diff
@@ -0,0 +1,347 @@
...
+In theory we can tell that `baz` cannot run but this would require graph traversal which could be complicated.
+For the moment we will not attempt this.
+
+One problem with this is that, even with required tasks, the workflow may never stall (depending on how the runahead limit is set).
+However the `waiting` tasks should at least be visible to the user and there should also be another visible cause (blocked or failed tasks) unless there is an error in the graph.
+
+## Cylc 7 compatibility
+
+Cylc 7 workflows will not be taking advantage of the new syntax which means there will only be required triggers.
+However there should always be corresponding suicide triggers where needed.
```
hjoliver, line 222: Note this section gets considerably easier in #129 because alternate branches don't get spawned at all (even as unsatisfied prerequisites) so they don't need to be suicided.

PR #129, opened 2021-07-19
New optional output syntax proposal

hjoliver:
Better alternative to PR #128

### DIFF
--- a/docs/proposal-new-output-syntax.md
+++ b/docs/proposal-new-output-syntax.md
@@ -1,107 +1,93 @@
-# Proposal: Optional Output Syntax
+# Proposal: Required and Optional Outputs
 
 Spawn-on-demand solves many problems but in its pure form it can't tell if the
-workflow has followed the "intended path" - to completion of the whole graph
-(say) vs going down a dead-end side branch. This is because all task outputs 
+workflow has followed the "intended path" - to completion of the whole graph,
+say, vs going down a dead-end side branch. This is because all task outputs 
 are effectively *optional*, and the scheduler just follows the flow wherever it
 leads at runtime.
 
 Cylc 7 could tell (albeit imperfectly!) if the intended path was followed 
-because all task outputs were effectively *required*. All possible paths are
+because all task outputs were effectively *required*: all paths were
 pre-spawned (to the next cycle point instance of each active task, roughly
-speaking). Users had to explicitly remove unused alternate paths with suicide
-triggers, and any remaining unsatisfied waiting tasks would indicate a problem
+speaking) and users had to explicitly remove unused paths with suicide
+triggers, so any remaining unsatisfied waiting tasks would indicate a problem
 by stalling the workflow.
 
 This document proposes a way to convey the intended path of execution to the
-Cylc 8 spawn-on-demand scheduler by making all task outputs **expected** by
+Cylc 8 spawn-on-demand scheduler by making all task outputs **required** by
 default and giving a new syntax to explicitly mark **optional** outputs. If a
-task finishes without completing all of its expected outputs it will be
-retained in the n=0 task pool and flagged as an **incomplete task** (if the
-task did not fail unexpectedly then this represents an error against the
-workflow configuration that was not recognized by the task job itself).
+task finishes without completing all of its required outputs it will be
+retained in the n=0 task pool and flagged as an **incomplete task** (often
+this will coincide with unexpected task failure; otherwise it implies that the
+task did not behave as the workflow configuration expected despite reporting
+success).
 
 *Note: optional triggers were considered and rejected as another way of
 potentially handling this problem - see cylc-admin#128*
 
-## The New Output Syntax
+## Output Syntax
 
-`foo:x` means `x` is an *expected output* of task `foo`. If a task finishes
-without completing all of its expected outputs, we will retain it in the task
-pool as an *incomplete task*.
+- `foo:x` means `x` is a *required output* of task `foo`
+  - if a task finishes without completing any required outputs, retain it in
+    the task pool as an *incomplete task*
 
-`foo?y` means `y` is an *optional output* of task `foo`. It is OK for a task to
-finish without completing its optional outputs.
+- `foo?y` means `y` is an *optional output* of task `foo`
+  - it is OK for a task to finish without completing optional outputs
 
 Abbreviated syntax for the success case:
+- `foo:` mean `foo:succeed` (success is required)
 - `foo?` means `foo?succeed` (success is optional)
-- `foo` and `foo:` mean `foo:succeed` (success is expected)
+- `foo` means `foo:succeed` UNLESS success is made optional elsewhere in the
+  graph with `foo?` or `foo?succeed` or `foo?fail`
   - this supports Cylc 7 graphs as well as being a sensible default
+  - we should probably give a validation warning here in case the user is
+    actually presuming contradictory success requirements for the same task
+    in different parts of the graph. For complete safety, if success is
+    optional use explicit syntax throughout.
 
-## Syntax Errors
-
-### succeed and fail
-
-The `succeed` and `fail` outputs are mutually exclusive opposites, therefore:
-
-- `succeed` and `fail` can't both be expected (or the task would be flagged as
-  incomplete no matter what)
-```
-# ERROR
-a
-a:fail
-```
-
-- `succeed` and `fail` can both be optional though (at a success/fail alternate
-  branching point)
-```
-# OK
-a?
-a?fail
-```
-
-- however it is OK to have one of `succeed` and `fail` expected and one
-  optional. This allows triggering off of the not-expected output at the same
-  time as the task is flagged as incomplete
-```
-# OK
-a
-a?fail => call_the_police
-```
-(Note this means we can't have *"`a` means `a:succeed` unless declared optional
-elsewhere"*.)
-
-### Custom Outputs
-
-Custom outputs must be either expected or optional, they can't be both
-```
-# ERROR
-a:x
-a?x
-```
-
-### QUESTION: submit-fail, submit, and start?
-
-It makes some sense to allow the `fail` output to be expected, but allowing
-`submit-fail` to be expected means treating a task as incomplete if it
-successfully submits. The only use case I've thought of
-is to make it somewhat easier (less cleanup before shutdown) to write
-functional tests that deliberately submit to a fake platform.
-
-Similarly, allowing `submit` and `start` to be optional equates to calling a
-task complete even though it didn't run at all, which seems equally perverse.
+### Interpretation in Trigger Expressions
 
 ```
-a:submit-fail  # ERROR or OK?
-a?submit  # ERROR or OK?
-a?start  # ERROR or OK?
+foo:x => bar
 ```
+... means trigger `bar` if `foo` completes output `x`; otherwise
+don't trigger `bar`, and flag `foo` as an incomplete task.
+
+```
+foo?x => bar
+```
+... means trigger `bar` if `foo` completes output `x`; otherwise
+don't trigger `bar`, but don't flag `foo` as an incomplete task either.
+
+### Caveats
+
+An output can't be both required and optional.
+
+Success should normally be required.
+
+Failure can be optional, in which case success must also be optional, because
+you might need alternate branches to handle:
+- a platform that isn't always available: `foo?submit-fail`
+- a task that you expect to fail sometimes: `foo?fail`
+
+Failure can't be required. Expecting a task to fail every time and treating
+it as incomplete if it succeeds would essentially amount to reversing normal 
+success and failure symantics. If you really have an executable that returns
+non-`0` whilst still doing what you need in the workflow, provide a wrapper
+that correctly returns success status.
+- `foo:fail  # ERROR`
+- `foo:submit-fail  # ERROR`
+
+Finally, optional `start` doesn't really make sense. At best it would be
+equivalent to `foo?submit`, so use that instead.
+- `foo?start  # ERROR`
+
 
 ## Path Branching
 
 ### Concurrent Paths 
 
-Concurrent paths can trigger off of expected outputs:
+Concurrent paths can trigger off of required outputs:
 ```
 a:x => b1
 a:y => b2
@@ -118,8 +104,8 @@ a?x => b1
 a?y => b2
 b1 | b2 => c  # join
 ```
-Note joining can trigger off of expected outputs, because the output is
-only "expected" if the task actually runs.
+Note joining can trigger off of required outputs, because the output is
+only "required" if the task actually runs.
 
 **Failure recovery** is a common example of this:
 ```
@@ -128,65 +114,32 @@ a?fail => b2  # failure path
 b1 | b2 => c  # join
 ```
 
-Note that an artificial dependency can be used to **ensure that at least one
-of the alternate paths gets taken**:
+Note that an artificial dependency can be used to *ensure that at least one
+of the alternate paths gets taken*:
 ```
 a?x => b1
 a?y => b2
 b1 | b2 => c  # join
 a => c # ARTIFICIAL DEPENDENCY
 ```
-The artificial trigger forces the unsatisfied prerequisites of `c` to be
-spawned, as waiting on `b1` or `b2` even if neither of the optional outputs 
-get completed. If the alternate paths don't actually need to join, `c` could be
-a dummy task.
+This forces `c` to be spawned as an unsatisfied prerequisite waiting on `b1` or
+`b2` even if neither branch runs. If the alternate paths don't actually need to
+join, `c` can be a dummy task.
 
-### QUESTION: Output Completion Expressions?
-
-In the previous example, an artificial dependency (possibly triggering a
-dummy task) is used to deliberately spawn an unsatisfied prerequisite in order
-to ensure that at least one of the alternate paths gets taken. However, that is
-not an ideal way to alert users to a problem:
-- the need to use an artificial dependency won't be obvious to users
-- an unsatisfied prerequisite can't be definitively identified as an error
-  until/unless the workflow stalls
-- and it is an indirect indicator: it does not highlight the actual task
-  that caused the problem
-
-An incomplete task is a direct and immediate indicator. So maybe we can allow
-users to say exactly when a task can be considered complete:
-
-```
-a:succeed(x|y) => c
-
-# or in abbreviated success form:
-a(x|y) => c
-
-# or if direct triggering of c is not needed:
-a(x|y)
-```
-This (or some variation on the suggested syntax) says that `a` is expected to
-succeed with at least one of `x` and `y` completed; otherwise it is incomplete
-even if it reports success.
+*See Appendix Output Completion Expressions for a possible follow-up that
+would allow us to flag `a` as incomplete here and avoid the artificial
+dependency.*
 
 ## Failed Tasks
 
 No special handling of failed tasks is required.
-```
-a => b  # success expected
-```
-If `a` fails it automatically gets retained in the `n=0` pool as an incomplete
-task (and will show as failed too).
+- if success is required, a failed task automatically gets retained in the task
+  pool as incomplete.
+- if success is optional (which would usually imply that optional failure is
+  explicitly handled), a failed task can be removed as complete
 
-Handled failed case:
-```
-a? => b
-a?fail => c
-```
-If `a` fails no problem, success is optional.
-
-No special syntax (e.g. `c:fail => NULL`) or suicide trigger (`c:fail => !c`)
-is needed for leaf tasks:
+No special syntax (e.g. `c:fail => NULL`) or suicide triggers (e.g. `c:fail =>
+!c`) are needed for leaf tasks either, if we don't care if they succeed or fail:
 ```
 a => b => c?  # success of c is optional
 ```
@@ -206,7 +159,7 @@ completed (i.e. keep them in n=0 as incomplete, or forget them).
 ```
 A:succeed-all => b
 ```
-- trigger b only if all members succeed; and all members are expected to
+- trigger b only if all members succeed; and all members are required to
   succeed
 
 ```
@@ -218,7 +171,7 @@ A?succeed-all => b
 ```
 A:succeed-any => b
 ```
-- trigger b if any member succeeds; but any member that runs is expected to
+- trigger b if any member succeeds; but any member that runs is required to
   succeed
 
 ```
@@ -231,7 +184,7 @@ A?succeed-any => b
 
 Incomplete tasks should be flagged immediately as a critical error.
 
-Most of the time they will be failed tasks that were expected to succeed.
+Most of the time they will be failed tasks that were required to succeed.
 Otherwise they represent an error against the workflow configuration that was
 not recognized by the task job itself. That could indicate a bug in the task
 (it reported success when it shouldn't have), or it could mean the workflow
@@ -247,7 +200,7 @@ Partially satisfied prerequisites should be made visible as an indication of a
 - Or they can result from a subtle workflow design error, where a task
   depends on several *alternate* branches at once:
   (e.g. when `a & b => c` should be `a | b = c`).
-- Or they can result from an upstream path being blocked when an expected
+- Or they can result from an upstream path being blocked when a required
   output doesn't get completed - which will now also result in an incomplete
   task.
 
@@ -291,7 +244,7 @@ We should also improve the visibility of stalled workflows and incomplete tasks
 ## New Event Handler
 
 We should add a new **incomplete task** event that is triggered whenever a task
-finishes without completing all of its expected outputs.
+finishes without completing all of its required outputs.
 
 ## Runahead limit
 
@@ -314,7 +267,7 @@ already have been in the task pool.
 ## Cylc 7 compatibility
 
 Cylc 7 workflows will not be using the new syntax which means they will only
-have expected outputs (but that's OK - Cylc 7 outputs **are** "expected" as
+have required outputs (but that's OK - Cylc 7 outputs **are** "required" as
 explained above).
 
 They also use suicide triggers to remove:
@@ -430,7 +383,7 @@ a?y => y1
 a?z => z1
 x1 | y1 | z1 => b
 
-a(x|y|z)  # a is expected to succeed with at least one of x, y, z
+a(x|y|z)  # a is required to succeed with at least one of x, y, z
   # OR
 a => b  # spawn b as a prerequisite that needs at least one branch to be satisfied
 ```
@@ -703,3 +656,65 @@ a?fail => r1 => b2 => r2
    [[b1, b2]]
        inherit = B
 ```
+
+# Appendix: Output Completion Expressions?
+
+A possible follow-up to this proposal.
+
+In an example above we used an artificial dependency (possibly triggering a
+dummy task) to deliberately generate a problem that can stall the workflow
+if neither branch runs in an alternate branch case:
+```
+a?x => b1
+a?y => b2
+b1 | b2 => c  # join
+a => c # ARTIFICIAL DEPENDENCY
+```
+
+The artificial dependency causes an unsatisfied `c` to be spawned even if
+neither branch runs.
+
+However, this isn't an ideal way to alert users to the problem
+- it might not occur to users to use an artificial dependency like this
+- unsatisfied prerequisites can't be definitively identified as errors
+  until/unless the workflow stalls
+- unsatisfied prerequisites are an indirect problem indicator: they do not
+  single out the task that actually caused the problem (which is that `a`
+should have generated either `x` or `y`)
+
+So maybe we can (correctly) identify `a` as incomplete instead:
+
+```
+a:succeed(x|y) => c
+
+# or in abbreviated success form:
+a(x|y) => c
+
+# or if direct triggering of c is not needed:
+a(x|y)
+```
+This (or some variation on the suggested syntax) says that `a` is required to
+succeed with at least one of `x` and `y` completed; otherwise it is incomplete
+even if it reports success.
+
+## Caveat?
+
+What if the optional outputs are not from the same task?
+```
+a => b1? & b2?
+b1? | b2? => c  # join
+a => c # ARTIFICIAL DEPENDENCY
+```
+In this case the artificial dependency is still needed to ensure that either
+`b1` or `b2` succeed.
+
+However, this example is arguably a workflow design error.
+
+If a task generates multiple outputs it is possible to say whether or not at
+least one output is expected - that is down to the internal functioning of the
+task. And if at least one is expected but it doesn't happen, then the task is
+incomplete.
+
+But if `b1` and `b2` above are both success-optional independent tasks, then
+surely two failures at once is a perfectly possible normal outcome, in which
+case it would make sense for the joining dependency to not be triggered.
--- a/docs/proposal-new-output-syntax.md
+++ b/docs/proposal-new-output-syntax.md
@@ -1,10 +1,10 @@
 # Proposal: Required and Optional Outputs
 
-Spawn-on-demand solves many problems but in its pure form it can't tell if the
-workflow has followed the "intended path" - to completion of the whole graph,
-say, vs going down a dead-end side branch. This is because all task outputs 
-are effectively *optional*, and the scheduler just follows the flow wherever it
-leads at runtime.
+Cylc 8 spawn-on-demand solves many problems but in its pure form it can't tell
+if the workflow has followed the "intended path" - to completion of the whole
+graph, say, vs going down a dead-end side branch. This is because all task
+outputs are effectively *optional*, and the scheduler just follows the flow
+wherever it leads at runtime.
 
 Cylc 7 could tell (albeit imperfectly!) if the intended path was followed,
 because it pre-spawned all possible paths (to the next cycle point instance of
@@ -12,13 +12,15 @@ each active task, roughly speaking) and expected users to remove unused paths
 with suicide triggers. Any remaining unsatisfied waiting tasks would stall 
 the workflow.
 
-This proposal introduces the concept of **task completion** with **required**
+This proposal introduces the concepts of **task completion**, and **required**
 and **optional** outputs, as a way to convey the intended path of execution to
-the Cylc 8 spawn-on-demand scheduler. *A task must complete all required outputs
-in order to be considered complete.* Otherwise it will be retained in the `n=0`
-task pool as incomplete, which will be flagged as error and (like partially
-satisfied prerequisites) will stall the scheduler if it has nothing else to do.
-Optional outputs do not contribute to task completion.
+the Cylc 8 scheduler. *Tasks that finish without completing all required
+outputs will be retained in the `n=0` task pool as incomplete tasks*, which
+can be flagged as error and (like partially satisfied prerequisites) will stall
+the scheduler if it has nothing else to do. Optional outputs do not contribute
+to task completion. Incomplete tasks will often be failed tasks that were
+required to succeed, otherwise they indicate that the task did not do what the
+graph expected despite reporting success.
 
 *Note: required and optional triggers were considered and rejected as another
 way of potentially handling this problem - see cylc-admin#128*
@@ -29,27 +31,26 @@ way of potentially handling this problem - see cylc-admin#128*
   - if `foo` finishes without completing `x`, retain it in n=0 as *incomplete*
 
 - `foo:y?` means `y` is an *optional output* of task `foo`
-  - it is OK for `foo` to finish without completing `x`
+  - it is OK for `foo` to finish without completing `y`
 
 A task is incomplete if:
-- job submission failed and `:submit/submit-fail` was not optional
-- or it finished executing and did not complete all required outputs
+- it finished executing without completing all required outputs
+- or if job submission failed and the `:submit` output was not optional
 
 Abbreviated syntax for the default success case:
 - `foo?` means `foo:succeed?` (success optional)
 - `foo` means `foo:succeed` (success required)
-  - `foo` can still fail, but it will be treated as incomplete if it does
 
 Examples in trigger expressions:
 ```
 # trigger bar if foo completes x; otherwise flag foo as incomplete
 foo:x => bar
 
-# trigger bar if foo completes x; either way, don't flag foo as incomplete
+# trigger bar if foo completes x; but either way don't flag foo as incomplete
 foo:x? => bar
 ```
 
-### Notes
+### Notes and Caveats
 
 The same output can't be both required and optional.
 
@@ -81,8 +82,8 @@ loop (use execution timeout) or if it never started running in the first place.
 The new optional output syntax cannot be used with family collective psuedo
 output triggers because it is difficult to interpret that in terms of the
 success status of each individual member, which is what we need.
-Instead we have different optional/required defaults for each collective
-pseudo-output. See [Family Triggers](#Family-Triggers) below. 
+Instead we set appropriate member defaults for each family trigger.
+See [Family Triggers](#Family-Triggers) below. 
 
 ## Path Branching
 
@@ -161,7 +162,7 @@ A:succeed-any => b  # member success (and hence failure) optional
 A:fail-all => b  # member failure required 
 A:fail-any => b  # member failure (and hence success) optional
 
-A:finish-all => b  # member success optional
+A:finish-all => b  # member success (and hence fail) ooptional
 A:finish-any => b  # member success (and hence fail) optional
 
 A:start-all => b  # N/A
--- a/docs/proposal-new-output-syntax.md
+++ b/docs/proposal-new-output-syntax.md
@@ -148,12 +148,12 @@ dummy task) is used to deliberately spawn an unsatisfied prerequisite in order
 to ensure that at least one of the alternate paths gets taken. However, that is
 not an ideal way to alert users to a problem:
 - the need to use an artificial dependency won't be obvious to users
-- an satisfied prerequisite can't be definitively identified as an error
+- an unsatisfied prerequisite can't be definitively identified as an error
   until/unless the workflow stalls
 - and it is an indirect indicator: it does not highlight the actual task
   that caused the problem
 
-An incomplete task is a direct and instant indicator. So maybe we can allow
+An incomplete task is a direct and immediate indicator. So maybe we can allow
 users to say exactly when a task can be considered complete:
 
 ```
@@ -252,7 +252,7 @@ Partially satisfied prerequisites should be made visible as an indication of a
   task.
 
 Note **output completion expressions** above will eliminate one source of
-partially satisfied prerequites.
+partially satisfied prerequisites.
 
 ### Can we tell if it is not possible for task prerequisities to be met?
 
--- a/docs/proposal-new-output-syntax.md
+++ b/docs/proposal-new-output-syntax.md
@@ -31,14 +31,14 @@ potentially handling this problem - see cylc-admin#128*
   - if a task finishes without completing any required outputs, retain it in
     the task pool as an *incomplete task*
 
-- `foo?y` means `y` is an *optional output* of task `foo`
+- `foo:y?` means `y` is an *optional output* of task `foo`
   - it is OK for a task to finish without completing optional outputs
 
 Abbreviated syntax for the success case:
 - `foo:` mean `foo:succeed` (success is required)
-- `foo?` means `foo?succeed` (success is optional)
+- `foo?` and `foo:?` mean `foo:succeed?` (success is optional)
 - `foo` means `foo:succeed` UNLESS success is made optional elsewhere in the
-  graph with `foo?` or `foo?succeed` or `foo?fail`
+  graph via `foo:succeed?` or `foo:fail?`
   - this supports Cylc 7 graphs as well as being a sensible default
   - we should probably give a validation warning here in case the user is
     actually presuming contradictory success requirements for the same task
@@ -54,7 +54,7 @@ foo:x => bar
 don't trigger `bar`, and flag `foo` as an incomplete task.
 
 ```
-foo?x => bar
+foo:x? => bar
 ```
 ... means trigger `bar` if `foo` completes output `x`; otherwise
 don't trigger `bar`, but don't flag `foo` as an incomplete task either.
@@ -67,8 +67,8 @@ Success should normally be required.
 
 Failure can be optional, in which case success must also be optional, because
 you might need alternate branches to handle:
-- a platform that isn't always available: `foo?submit-fail`
-- a task that you expect to fail sometimes: `foo?fail`
+- a platform that isn't always available: `foo:submit-fail?`
+- a task that you expect to fail sometimes: `foo:fail?`
 
 Failure can't be required. Expecting a task to fail every time and treating
 it as incomplete if it succeeds would essentially amount to reversing normal 
@@ -79,7 +79,7 @@ that correctly returns success status.
 - `foo:submit-fail  # ERROR`
 
 Finally, optional `start` doesn't really make sense. At best it would be
-equivalent to `foo?submit`, so use that instead.
+equivalent to `foo:submit?`, so use that instead.
 - `foo?start  # ERROR`
 
 
@@ -100,8 +100,8 @@ Alternate paths must branch off of optional outputs, and must be joined with
 OR triggers because only one side or the other will run:
 
 ```
-a?x => b1
-a?y => b2
+a:x? => b1
+a:y? => b2
 b1 | b2 => c  # join
 ```
 Note joining can trigger off of required outputs, because the output is
@@ -110,15 +110,15 @@ only "required" if the task actually runs.
 **Failure recovery** is a common example of this:
 ```
 a? => b1  # success path
-a?fail => b2  # failure path
+a:fail? => b2  # failure path
 b1 | b2 => c  # join
 ```
 
 Note that an artificial dependency can be used to *ensure that at least one
 of the alternate paths gets taken*:
 ```
-a?x => b1
-a?y => b2
+a:x? => b1
+a:y? => b2
 b1 | b2 => c  # join
 a => c # ARTIFICIAL DEPENDENCY
 ```
@@ -163,7 +163,7 @@ A:succeed-all => b
   succeed
 
 ```
-A?succeed-all => b
+A:succeed-all? => b
 ```
 - trigger b only if all members succeed; but it's OK if they fail (maybe the
   failed path is handled elsewhere)
@@ -175,7 +175,7 @@ A:succeed-any => b
   succeed
 
 ```
-A?succeed-any => b
+A:succeed-any? => b
 ```
 - trigger b if any member succeeds; but it's OK if any or all members fail
 
@@ -279,7 +279,7 @@ For example, in this Cylc 8 graph the `c1 => c2` branch does not get spawned
 at all unless `a` fails:
 ```
 a? => b1 => b2
-a?fail => c1 => c2
+a:fail? => c1 => c2
 b2 | c2 => d
 ```
 
@@ -336,7 +336,7 @@ prerequisite:
         R1 = """
             a => b => c => d
             a => check-d? => d
-            check-d?fail => !d
+            check-d:fail? => !d
         """
 [runtime]
    ...
@@ -352,7 +352,7 @@ handle this properly:
         R1 = """
             a => b => c
             a => check-d
-            check-d?fail => { }  # or just "check-d?fail"
+            check-d:fail? => { }  # or just "check-d:fail?"
             check-d? & c => { d }
         """
 [runtime]
@@ -378,9 +378,9 @@ break the workflow if more than one completes.
 Using the new syntax this would become:
 ```
 # Cylc 8
-a?x => x1
-a?y => y1
-a?z => z1
+a:x? => x1
+a:y? => y1
+a:z? => z1
 x1 | y1 | z1 => b
 
 a(x|y|z)  # a is required to succeed with at least one of x, y, z
@@ -449,7 +449,7 @@ cylc-flow#3282?
     [[graph]]
         R1 = """
              foo? => bar => qux
-             foo?fail => baz => qux
+             foo:fail? => baz => qux
              """
 [runtime]
     [[foo,bar,baz,qux]]
@@ -469,7 +469,7 @@ partially satisfied because it also depends on the other path.
         P1 = """
              model[-P1] => model => archive
              archive[-P1]? => archive
-             archive?fail => recover
+             archive:fail? => recover
              """
 [runtime]
     [[model,archive,recover]]
@@ -536,7 +536,7 @@ z:
 
 ```
 a? => b  
-a?fail => c 
+a:fail? => c 
 b | c => z
 ```
 
@@ -545,7 +545,7 @@ each alternate branch.
 ```
 # ERROR: this does not result in separate `x`'s on each branch
 a? => b => x => m => z
-a?fail => c => x => n => z
+a:fail? => c => x => n => z
 m | n => z
 ```
 
@@ -562,7 +562,7 @@ represent the same x job content, in each branch:
 
 ```
 a? => b => x1 => m => z
-a?fail => c => x2 => n => z
+a:fail? => c => x2 => n => z
 m | n => z
 ...
 [runtime]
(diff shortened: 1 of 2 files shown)


oliver-sanders:
Just to check, this proposal keeps the `=>` operator behaving the same way as in 8.0b2, however, adds an extra step on task/output completion to check that the outputs produced match. If this check fails we block spawning downstream and mark the upstream as incomplete.

oliver-sanders:
Awkward example, how would this behave if `a` produces `n` but not `o`?

```
# success pathway
a?m => b => c

# fail pathway
a?n => x => y => z
a:o => z
```

dpmatthews:
> Just to check, this proposal keeps the `=>` operator behaving the same way as in 8.0b2,

I think partially satisfied dependencies are assumed to stall the workflow in this proposal which is not true in 8.0b2.

hjoliver:
> > Just to check, this proposal keeps the `=>` operator behaving the same way as in 8.0b2,
> 
> I think partially satisfied dependencies are assumed to stall the workflow in this proposal which is not true in 8.0b2.

Correct. 

Mind you, we could have had partially satisfied prereqs stall 8.0b2 too.  What we can't do without one form or another of the new syntax is stall on unsatisfied single prerequisites (hence my main beef with stalling at 8.0b2 - inconsistent treatment of single and multi-prerequisite cases).

hjoliver:
> Just to check, this proposal keeps the `=>` operator behaving the same way as in 8.0b2, however, adds an extra step on task/output completion to check that the outputs produced match. If this check fails we block spawning downstream and mark the upstream as incomplete.

Yes, although I'd phrase it slightly differently: we only spawn on outputs, but when a task finishes we check to see if all of its expected/required outputs were completed. If not, we mark the task as incomplete and keep in alive in the n=0 pool.  No need to "block spawning downstream" because the output that we would have spawned on was not completed.

hjoliver:
> Awkward example, how would this behave if `a` produces `n` but not `o`?
> 
> ```
> # success pathway
> a?m => b => c
> 
> # fail pathway
> a?n => x => y => z
> a:o => z
> ```

- `x` and `y` would run, and  `z` would be spawned as waiting with its dependence on `a:o` unsatisfied
- `a` would be marked as incomplete, because `a:o` is an expected output

So we could probably just highlight the incomplete `a` as the problem, and leave the partially satisfied `z` hidden.  Fixing `a` will automatically fix `z`.

oliver-sanders:
I've started collecting a list of examples, writing them out with 7/8.0b2/#128/#129 logic.

I have hit one (non-pathological) example where this proposal may encounter a hitch.

This is a real-world example sent to us via user-support, I can vouch that it is a sane use case...

# Parallel Pipelines That Can Finish Early

In this example we have the following pipeline:

```
a<x> => b<x> => d<x>
b<x> => c => d<x>
```

abcd


With the condition that if any of the `a<x>` tasks we skip the rest of the tasks in the `<x>` pipeline.

I.E. if `a2` fails, we skip the pipe `a2 => b2 => d2` but continue to run it for `a1` and `a3`.

The following examples rely on the inheritance:

```ini
[runtime]
    [[<x>]]
    [[a<x>, b<x>, c<x>]]
        inherit = <x>
```

## Cylc 7

```
a<x> => b<x>
a<x>:fail | b<x> => c => d<x>
a<x>:fail => !<x>
```

## Cylc 8.0b2

```
a<x> => b<x> => d<x>
a<x>:fail | b<x> => c
c => d<x>
```

## #128

```ini
a<x> ?=> b<x> => d<x>
a<x>:fail | b<x> => c
c ?=> d<x>  # somewhat counter-intuitive but functional reverse use of an optional trigger
```

## #129

```ini
a<x>? => b<x> => d<x>
a<x>:fail | b<x> => c
c => d<x>  # ERROR? Would this cause all d<x> to be spawned irrespective of the a<x> => b<x> => d<x> pipeline?
```

Under 8.0b2 the `c => d<x>` dependency didn't cause any issues, with #128 we maintain that behaviour using `?=>`.

dpmatthews:
I've wondering if we can combine the optional/required outputs with an optional trigger syntax, e.g.:
```
a<x>? => b<x> => d<x>
a<x>:fail | b<x> => c
c =?> d<x>
```
where `c =?> d<x>` implies: if `c` succeeds we spawn `d<x>` if all it's prerequisites are met.

dpmatthews:
Is there a problem with family triggers?
If tasks `a1`,`a2`,`a3` belong to the family `A`, how do we handle `succeed-any`?
`A?succeed-any` is presumably equivalent to `a1? | b1? | c1?`.
However, `A:succeed-any` can't be equivalent to `a1 | b1 | c1` since that would force them all to succeed.

hjoliver:
> I have hit one (non-pathological) example where this proposal may encounter a hitch.

This example is actually pathological in a similar sense to example 4 in the proposal.  The problem is that the task `c` needs to have different outputs/triggers depending on upstream events, but Cylc 7/8 graph parsing can't handle that - it combines *all* triggers for all graph paths that the task appears in.

Once again, it works fine with future Cylc 9 dynamic sub-graphs:
```
# Cylc 9 initial graph
a<x> => { b<x> => c => d<x>}
a<x>:fail => {c}
```
(I'm ignoring `b<x> => d<x>` as redundant, but it doesn't change anything)
So, `c` needs different dependencies according to which `a<x>` tasks succeed or fail.  E.g. if `a1` fails, we should get this at runtime:
```
# Cylc 9 dynamic graph, when a1 fails
a2 => b2 => c => d2
a3 => b3 => c => d3
a1:fail => c
```
... and `c` will (correctly) not spawn `d1` :tada:

It works at Cylc 7 only because suicide triggers are a big blunt instrument that can take out incorrectly spawned tasks.

The right way to do this at Cylc 8 is again to use two logical `c` tasks (with different dependencies) to represent the same `c` job:
```
# Cylc 8
a<x>? => b<x> => cA => d<x>
a<x>?fail => cB
...
[runtime]
   [[C]]
   [[cA, cB]]
      inherit = C
```
The difference from example 4 is that here both cA and CB may run at once as they're not on strictly alternate branches. So there could be two `c` tasks on the same running the same underlying job, but that's a very small inefficiency.

However, it may still be useful to invent an intuitive special syntax to handle the incorrect dependencies without using different logical tasks...

hjoliver:
This has been a useful exercise, to clarify exactly what we're still missing out on by not supporting dynamic subgraphs yet.

- Cylc 8 now only spawns new tasks "on demand" :+1: 
   - so for alternate branches with different tasks on each branch (which thankfully is usually the case) the unused branch does not get spawned so we no longer have to remove it with suicide triggers 
- HOWEVER we still harvest prerequisites and outputs indiscriminately from every place in the graph, so *if the same task appears in alternate branches with different dependencies*, every instance of that task will get *all* of those dependencies. :-1: 
  - affected tasks may incorrectly get stuck as waiting, because they have picked up extra prerequisites that can only be satisified on another branch
  - or they may incorrectlyt spawn extra children, because they have picked up extra outputs from another branch

hjoliver:
(Family triggers TBD...)

dpmatthews:
Syntax: I don't think we should use `?=>` because it's too difficult to distinguish between `a? => b` and `a ?=> b`. Suggestion: `=>?`.

Description: I still think required and optional triggers is a reasonable description.

`=>` is a required trigger. Child tasks of this trigger are required to run (once their prerequisites have been satisfied) in order for the workflow to complete successfully. If there is nothing else to run and any required tasks remain with partially satisfied prerequisites then the workflow will stall. Required tasks with partially satisfied prerequisites also count towards the runahead limit. If it is not possible for a required task to run (because it is not possible for it's prerequisites to be met) then the task will be flagged as "blocked".

`=>?` is a optional trigger. Child tasks of this trigger will run as normal if fully satisfied, but will be ignored if they remain partially satisfied. Optional tasks with partially satisfied prerequisites do not cause stalls or count towards the runahead limit.

hjoliver:
@dpmatthews - OK agreed on syntax, name, and description - that makes perfect sense :+1:

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,211 @@
...
+Handled failed case:
+```
+a? => b
+a?fail => c
+```
+If `a` fails no problem, success was optional.
+
+Bonus, no additional special syntax (e.g. `c:fail => NULL` or `c:fail => !c`) is needed to handle leaf tasks:
+```
+a => b => c?  # success of c is optional
```
dpmatthews, line 95: Well, it is sort of special syntax because RHS triggers usually don't mean anything (I didn't realise they were allowed).

What's the logic going to be? Success is assumed to be required unless there is a `succeed` or `fail` trigger?

```
a => c
b => c? # Success of c is optional
# Also allowed:
b => c?succeed
b => c?fail
```
```
a => b
b?fail => c # Success of b is optional
```
```
a => b
b:fail => c # Failure of b is required!
```

oliver-sanders: Yes, `c?` is shorthand for `c?succeeded` similar to `c` being shorthand for `c:succeeded`.

hjoliver: @dpmatthews remember `?succeed` is an output, not a trigger.   So `b => c?` means: `c triggers off of b; then when c runs, success is not a required output* (so if it does not succeed, we don't have to keep it in n=0 as an incomplete/blocked task).

hjoliver: > ```
> a => b
> b:fail => c # Failure of b is required!
```

Yes, fail triggers should almost always be this:
```
b?fail => c  # failure is "optional" but if it happens trigger c
```
But note will now be able to have tasks that are expected to fail - which might occasionally be useful!

Comment (oliver-sanders) [reply docs/proposal-new-output-syntax.md:95]: > we don't have to keep it in n=0 as an incomplete/blocked task

I think Dave would like to keep the tasks in n=0 to make it easier for the user to fix the issue (retrigger, trigger-edit, artificially satisfy outputs)?

Comment (hjoliver) [reply docs/proposal-new-output-syntax.md:95]: In that case, use `b => c` (and keep `c` in n=0 if it fails) instead of `b => c?`.

Comment (hjoliver) [reply docs/proposal-new-output-syntax.md:95]: Instead of "keep unhandled failed tasks in n=0" we now have "keep incomplete tasks in n=0", which automatically includes failed tasks whose success is required (`b => c`)

Comment (dpmatthews) [reply docs/proposal-new-output-syntax.md:95]: > `?succeed` is an output, not a trigger. So `b => c?` means: _c triggers off of b; then when c runs, success is not a required output_ (so if it does not succeed, we don't have to keep it in n=0 as an incomplete/blocked task).

Yes, sorry - didn't explain myself very well. Presumably you agee that `a => b` doesn't mean that success of `b` is required (even though `b` is shorthand for `b:succeed`)? See examples above.

If there is neither a `fail` output nor a `succeed` output defined then `succeed` is assumed to be required.

What about `a => b:succeed`. Does that make success of `b` required? Presumably yes (although I can't see why you'd need to use this).

Comment (hjoliver) [reply docs/proposal-new-output-syntax.md:95]: Sorry, I'm confused about what you're saying here exactly. 

Does this statement:
> Presumably you agree that `a => b` doesn't mean that success of `b` is required?

not contradict this one?:
> If there is neither a `fail` output nor a `succeed` output defined then `succeed` is assumed to be required.

I think `a => b` **does** imply success of `b` is required.  That's exactly why we can keep the failed `b` in n=0 as an incomplete task.

Comment (hjoliver) [reply docs/proposal-new-output-syntax.md:95]: Maybe you're worried about this:
```
a => b
b:fail => c # Failure of b is required!
```
which says both success and failure of b are required.   That's an error because `:succeed` and `:fail` are mutually exclusive outputs, so we should make it fail validation.

Comment (dpmatthews) [reply docs/proposal-new-output-syntax.md:95]: So, if I understand correctly, I assume that means this is an error:

```
a => b
b? => c
b?fail => d
```
i.e. `:succeed` and `?fail` are mutually exclusive outputs

So this is also an error:
```
a => b?   # Should be a => b:fail
b:fail => c
```

What about this?
```
a => c
b => c?
```
Are `:succeed` and `?succeed` mutually exclusive?

> That's an error because `:succeed` and `:fail` are mutually exclusive outputs, so we should make it fail validation.

Yes, I think it would be essential to detect graphing errors at validation (too easy to make mistakes otherwise).
However we can't do this for Cylc 7 workflows so we'd need to do something to detect Cylc 8 workflows (based on use of newer syntax?).

Comment (hjoliver) [reply docs/proposal-new-output-syntax.md:95]: > Are `:succeed` and `?succeed` mutually exclusive?

Yes, because the "output type" (for want of a better name?) refers to the **output** not to the trigger, so it is not possible for a single output to be both required and not-required.  So this is definitely an error:
```
# ERROR
a:succeed => b
a?succeed => c
```
Similarly:
```
# ERROR
a => z:succeed
b => z?succeed
```

If "output status" (optional vs required) is not specified we have two options:
1. it means success is required, or
2. it means success is required *unless

Comment (hjoliver) [reply docs/proposal-new-output-syntax.md:95]: So your first example is only an error if we `b` **always** means `b:succeed` (as opposed to: assume `:succeed` unless told otherwise):
```
a => b
b? => c
b?fail => d
```

Your second example is an error, because it doesn't make sense for failure to be required at the same time as success is optional.  Succeed and fail are mutually exclusive so either one has to be required, or both have to be optional (but you don't have to write both optional cases out if you only trigger off of one of them).

```
a => b?   # Should be a => b:fail
b:fail => c
```
And your final example is a simpler version of the first. It's an error unless we go for the less prescriptive "assume success required unless told otherwise" option:
```
a => c
b => c?
```

Comment (hjoliver) [reply docs/proposal-new-output-syntax.md:95]: The vast majority of tasks will be "success required" of course, so users really only have to use this syntax at alternate path branch points and leaf tasks where success/fail doesn't matter.

Comment (hjoliver) [reply docs/proposal-new-output-syntax.md:95]: I vote for the more flexible option: assume success is required unless told otherwise.  That way, we don't have to explicitly use a `?` every time an affected task appears at the end of a graph line.

**Comment by hjoliver** [comment]
## Family triggers


Conceivably we could try to interpret the output signifier in terms of a collective pseudo-output for the family as whole, but that gets really hard to understand :exploding_head: 

Fortunately, I think the output type signifier can simply be applied directly to all family members, for all family trigger types. That's easy to understand, and it makes sense because the signifier doesn't affect triggering, it only affects what we do with family members if the outputs in question are not completed (i.e. keep them in n=0 as incomplete, or forget about them).


```
A:succeed-all => b
```
- trigger `b` only if all members complete; and all members are expected to succeed

```
A?succeed-all => b
```
- trigger `b` only if all members succeed; but it's OK if they fail (maybe the failed path is handled elsewhere)

```
A:succeed-any => b
```
- trigger `b` if any member succeeds; but if members run they are expected to succeed
  
```
A?succeed-any => b
```
- trigger `b` if any member succeed; but it's OK if other members fail

File: docs/proposal-new-output-syntax.md (hunk 1 above)
Comment (dpmatthews) [reply docs/proposal-new-output-syntax.md:95]: Whichever option we choose, I think we have to _assume success is required unless it's explicitly made optional elsewhere in the graph_ to deal with cases like `a => b:x => c` or `a => b:start =? c`. In these cases we need to assume success is required.
Also note that any of `?succeed`, `:fail`, `?fail` will make success optional.

I tend to think it will be easier if we always treat `b` as a shorthand for `b:succeed` rather than trying to explain the exception. This is likely to lead to more graphing errors but that's OK if we can fail at validation (and if we detect a Cylc 7 workflow we can still issue warnings since these will help identify where the graph needs changing).

**Comment by wxtim** [comment]
(Possible relevant I've spotted whilst trying to create Platforms test workflows)

As far as I can tell submit-fail is not treated as a handled failure. Consider

```ini
[scheduling]
    initial cycle point = 1
    [[graph]]
        R1 = """
            bad:fail => good
            ugly:submit-fail => good
        """

[runtime]
    [[root]]
        script = true
    [[good]]
    [[bad]]
        script = false
    [[ugly]]
        platform = blantantly_should_not_work
```

will result in :
```bash
2021-07-27T14:50:35+01:00 WARNING - Workflow stalled with unhandled failed tasks:
	* ugly.1 (submit-failed)
```

hjoliver:
> (Possible relevant I've spotted whilst trying to create Platforms test workflows)
> 
> As far as I can tell submit-fail is not treated as a handled failure. Consider ...

Looks like you've identified an omission on master. The distinction between handled-fail and not-handled-fail isn't needed under this proposal, but I guess I should still address submit(-failed) explicitly in the document. 

The question is, should we allow `submitted` be an optional output? (and allow `submit-falied` to be expected?)  My initial thought was no, but it might make some functional tests a bit easier - those that contrive to get a failed job submission.

wxtim:
>  those that contrive to get a failed job submission.

Hmmm... You'll never guess what I've been doing...

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,702 @@
...
+a?x
+```
+
+### QUESTION: submit-fail, submit, and start?
+
+It makes some sense to allow the `fail` output to be expected, but allowing
+`submit-fail` to be expected means treating a task as incomplete if it
+successfully submits. The only use case I've thought of
+is to make it somewhat easier (less cleanup before shutdown) to write
+functional tests that deliberately submit to a fake platform.
```
wxtim, line 89: Can live without it. I'm not sure that its enough of a use case.

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,702 @@
...
+retained in the n=0 task pool and flagged as an **incomplete task** (if the
+task did not fail unexpectedly then this represents an error against the
+workflow configuration that was not recognized by the task job itself).
+
+*Note: optional triggers were considered and rejected as another way of
+potentially handling this problem - see cylc-admin#128*
+
+## The New Output Syntax
+
+`foo:x` means `x` is an *expected output* of task `foo`. If a task finishes
```
wxtim, line 29: I'm a bit sleep befuddled/under-caffinated this morning but I'm finding this easier to understand if I `s/expected/required/`

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,702 @@
...
+
+- `succeed` and `fail` can't both be expected (or the task would be flagged as
+  incomplete no matter what)
+```
+# ERROR
+a
+a:fail
+```
+
+- `succeed` and `fail` can both be optional though (at a success/fail alternate
```
wxtim, line 55: If a task has only optional outputs, what is considered completion? Any of the options being triggered?

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,702 @@
...
+
+### QUESTION: submit-fail, submit, and start?
+
+It makes some sense to allow the `fail` output to be expected, but allowing
+`submit-fail` to be expected means treating a task as incomplete if it
+successfully submits. The only use case I've thought of
+is to make it somewhat easier (less cleanup before shutdown) to write
+functional tests that deliberately submit to a fake platform.
+
+Similarly, allowing `submit` and `start` to be optional equates to calling a
```
wxtim, line 91: I accept that if submit-fail is expected then submit must be optional.
Does it follow that if submit is optional that submit-fail must be required/expected? Surely both could be optional?
If I have understood correctly, then `a?submit` and `a?start` are wierd, but not wrong in the way that `a:submit-fail` _is_.

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,702 @@
...
+b1 | b2 => c  # join
+```
+
+Note that an artificial dependency can be used to **ensure that at least one
+of the alternate paths gets taken**:
+```
+a?x => b1
+a?y => b2
+b1 | b2 => c  # join
+a => c # ARTIFICIAL DEPENDENCY
```
wxtim, line 137: Do we have conventions for graphing this sort of thing? I'm finding that I have a notebook full of increasingly alchemical looking drawings...

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,702 @@
...
+a dummy task.
+
+### QUESTION: Output Completion Expressions?
+
+In the previous example, an artificial dependency (possibly triggering a
+dummy task) is used to deliberately spawn an unsatisfied prerequisite in order
+to ensure that at least one of the alternate paths gets taken. However, that is
+not an ideal way to alert users to a problem:
+- the need to use an artificial dependency won't be obvious to users
+- an satisfied prerequisite can't be definitively identified as an error
```
wxtim, line 151: an unsatisfied?

docs/proposal-new-output-syntax.md
hjoliver: I considered "required" too, but it sounds more coercive. We're not actually doing anything to "require" that the output gets completed, but the graph config nevertheless "expects" it to be completed (and gets a surprise if it isn't).

docs/proposal-new-output-syntax.md
hjoliver: If all of a task's outputs are optional (including succeed and fail) then however it *finishes* (succeed or fail) it is complete. It might be easier to think of it as being incomplete only if some expected outputs were not completed.

docs/proposal-new-output-syntax.md
hjoliver: No, but it might be good to get more information into our visualizations, beyond graph node means task, and graph edge means a dependency of some kind.  I'm not sure I like conditionals as graph nodes though ... too much like a task called "or"?

docs/proposal-new-output-syntax.md
hjoliver: Agreed.

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,705 @@
...
+  branching point)
+```
+# OK
+a?
+a?fail
+```
+
+- however it is OK to have one of `succeed` and `fail` expected and one
+  optional. This allows triggering off of the not-expected output at the same
+  time as the task is flagged as incomplete
```
dpmatthews, line 65: Do we really want / need this? I think it's going to be much easier to explain to users and much easier to detect issues if we say that `succeed` and `fail` must both be optional if one is. I think this is correct in terms of the workflow definition. If action needs to be taken as a result of a task not working correctly (produced the expected outputs) that isn't that what event triggers are for?

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,705 @@
...
+# OK
+a
+a?fail => call_the_police
+```
+(Note this means we can't have *"`a` means `a:succeed` unless declared optional
+elsewhere"*.)
+
+### Custom Outputs
+
+Custom outputs must be either expected or optional, they can't be both
```
dpmatthews, line 76: Isn't that true for all outputs (succeed, fail, etc)?

docs/proposal-new-output-syntax.md
dpmatthews: Does it really make sense to allow `fail` to be expected? Retry is designed around the fact that tasks are expected to succeed. If a task which is expected to fail succeeds do we need to be able to configure it to retry until it fails? That seems crazy. I think I'd prefer us to say that `fail` (and `submit-fail`) cannot be expected.

docs/proposal-new-output-syntax.md
dpmatthews: I think allowing `submit` to be optional is fine. You might want to submit a task to a platform which isn't always available.

I can't see how `start` being optional would make sense (it would effectively imply `submit` is optional).

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,705 @@
...
+to ensure that at least one of the alternate paths gets taken. However, that is
+not an ideal way to alert users to a problem:
+- the need to use an artificial dependency won't be obvious to users
+- an unsatisfied prerequisite can't be definitively identified as an error
+  until/unless the workflow stalls
+- and it is an indirect indicator: it does not highlight the actual task
+  that caused the problem
+
+An incomplete task is a direct and immediate indicator. So maybe we can allow
+users to say exactly when a task can be considered complete:
```
dpmatthews, line 157: This looks complicated and only helps if you're dealing with multiple outputs from the same task.
For example:
```
a => b1? & b2?
b1? | b2? => c  # join
a => c # ARTIFICIAL DEPENDENCY
```
In this case the artificial dependency is still needed to ensure that either `b1` or `b2` succeed.

docs/proposal-new-output-syntax.md
dpmatthews: I don't think the proposal covers this discussion yet?
I think we have to assume `succeed` is required unless it's explicitly made optional elsewhere in the graph (via `?succeed` or `?fail`).
Similarly for `submit`.
And `a => b` implies `succeed` is expected for `b` regardless of what else is in the graph.

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,705 @@
...
+because the suicide trigger that spawns it also satisfies its only suicide
+prerequisite.
+
+It doesn't matter if a task still has unsatisfied normal prerequisites when it
+suicides. The other triggers can't spawn it again in the same flow thanks to
+our conditional reflow prevention mechanism.
+
+For the Cylc 7 graph above to work with the new optional output syntax we just
+need to infer that `a:succeed` and `a:fail` are optional, to avoid `a` ending
+up marked as incomplete both ways.
```
dpmatthews, line 357: I'm not sure what you mean here. My assumption is that we have to treat `succeed` and `fail` as required so that `a` remains in the task pool until such time as it is removed via a suicide trigger.

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,705 @@
...
+## Syntax Errors
+
+### succeed and fail
+
+The `succeed` and `fail` outputs are mutually exclusive opposites, therefore:
+
+- `succeed` and `fail` can't both be expected (or the task would be flagged as
+  incomplete no matter what)
+```
+# ERROR
```
dpmatthews, line 50: What does error mean in this case - validation error?
We can do that for a cylc 8 graph (`flow.cylc`) but not a cylc 7 graph (`suite.rc`)?

docs/proposal-new-output-syntax.md
hjoliver: Fair point, I'll go with that.

docs/proposal-new-output-syntax.md
hjoliver: Fair enough. I suppose if a user really has an executable that they expect not to exit with `0` whilst still doing what they need it to do, they can provide a simple wrapper to turn that into success for the workflow.

docs/proposal-new-output-syntax.md
hjoliver: Actually I've changed my mind on this!  If an output is *required* it can't be *optional*, but if it is optional we can *expect* that it might not be completed.

docs/proposal-new-output-syntax.md
hjoliver: Agreed.

docs/proposal-new-output-syntax.md
hjoliver: Yes, updated.

docs/proposal-new-output-syntax.md
hjoliver: (no longer relevant after to other changes)

docs/proposal-new-output-syntax.md
hjoliver: OK, I'll move that section to an appendix, as a possible follow-up.  And I'll note your example there too.

hjoliver: Moved to an appendix as a possible follow-up.  But on reflection I don't think your example makes much sense - see the new "caveat?" sub-section in the proposal.

docs/proposal-new-output-syntax.md
hjoliver: No, I don't think so.  Suicide triggers necessarily work differently under spawn-on-demand, but that doesn't matter so long as we get the same outcome.  Here's a simpler example:

```
# Native Cylc 8
a? => b
a?fail => c
```
... `b` only gets spawned if `a` succeeds; and `c` only gets spawned if `a` fails

```
# Cylc 7
a => b
a:fail => c
a:fail => !b
a => !c
c => !a 
```
...both `b` and `c` are pre-spawned. If `a` fails, it gets removed later on by suicide. And `b` or `c` get removed immediately, depending on the final state of `a`.

If we run the Cylc 7 case with Cylc 8:
- we can safely infer that success (and failure) of `a` is optional, because the alternative is impossible
   - then we don't end up with `a` incomplete, whichever outcome occurs
- later, if `a` fails (say), the suicide triggers will:
- spawn `b`, which will immediately be removed again because it's suicide prerequisite is satisfied by the trigger that spawned it
-  be blocked from spawning `a` because `a` was already spawned earlier in this flow

docs/proposal-new-output-syntax.md
hjoliver: > I think we have to assume succeed is required unless it's explicitly made optional elsewhere in the graph (via `?succeed` or `?fail`).
Similarly for submit.

Agreed, I'll update the document to make that clear.

>  And `a => b` implies `succeed` is expected for `b` regardless of what else is in the graph.

I don't understand this. Presumably you're not saying that `a` has to succeed in order to trigger `b`, which obviously true regardless of whether `a:succeed` is a required or optional output.  So you must be saying that `a => b` means `a => b:succeed` even if `b?succeed` or `b?fail` appear elsewhere in the graph?  But that contradicts the general rule above. :confounded:

docs/proposal-new-output-syntax.md
hjoliver: What you're suggesting probably is an option too, because the suicide trigger would still remove if the incomplete task.  But it would be better to avoid the incomplete task in the first place, if we can.

hjoliver: Even @oliver-sanders nasty Cylc 7 aborted pipeline case above runs correctly on current master:
```
a1 => b1
a1:fail | b1 => c => d1
a1:fail & c => !a1 & !b1 & !d1
a2 => b2
a2:fail | b2 => c => d2
a2:fail & c => !a2 & !b2 & !d2
```
Result with `a2` failing:
```console
$ cylc cat-log d8/run1 | grep trig
2021-07-29T17:19:07+12:00 INFO - [a1.1] -triggered off []
2021-07-29T17:19:07+12:00 INFO - [a2.1] -triggered off []
2021-07-29T17:19:15+12:00 INFO - [b1.1] -triggered off ['a1.1']
2021-07-29T17:19:24+12:00 INFO - [c.1] -triggered off ['a2.1', 'b1.1']
2021-07-29T17:19:33+12:00 INFO - [d1.1] -triggered off ['c.1']
```
On master all outputs are already optional, so we just need to infer `a1?succeed` etc. to get the same result with the new syntax.

MetRonnie:
#### Syntax

I think that if the `?` goes after the output, that would make more sense.

| Original proposal    | My suggestion         |
| -------------------- | --------------------- |
| `foo?x => bar`       | `foo:x? => bar`       |
| `foo?fail => bar`    | `foo:fail? => bar`    |
| `foo?succeed => bar` | `foo:succeed? => bar` |
| `foo? => bar`        | `foo? => bar` (same)  |

hjoliver:
> I think that if the ? goes after the output, that would make more sense.

Brilliant idea :tada: 

Let's see if anyone disagrees, but I like it!

oliver-sanders:
I would prefer that, I assumed you were going minimal avoiding an extra character.

hjoliver:
Syntax change made, `foo?x` is now `foo:x?` etc.

docs/proposal-new-output-syntax.md
dpmatthews: It doesn't contradict. The general rule only applies in cases where success or failure hasn't been specified as expected or optional. It's required for cases like `a => b:start => c` (with no other reference to `b` in the graph) - `succeed` hasn't been made expected or optional in this case for `b` so we assume it is expected.

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,725 @@
...
+it as incomplete if it succeeds would essentially amount to reversing normal 
+success and failure symantics. If you really have an executable that returns
+non-`0` whilst still doing what you need in the workflow, provide a wrapper
+that correctly returns success status.
+- `foo:fail  # ERROR`
+- `foo:submit-fail  # ERROR`
+
+Finally, optional `start` doesn't really make sense. At best it would be
+equivalent to `foo:submit?`, so use that instead.
+- `foo?start  # ERROR`
```
MetRonnie, line 83: One slipped through the net
```suggestion
- `foo:start?  # ERROR`
```

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,725 @@
...
+[scheduling]
+    cycling mode = integer
+    initial cycle point = 1
+    [[graph]]
+        P1 = """
+             bar[-P1] => foo => bar & baz
+             baz[-P1] => baz
+             """
+```
+Result: `baz.3` stuck as partially satisfied waiting on `bar.2`
```
dpmatthews, line 496: Typo: `baz.2` (not `bar.2`)

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,725 @@
...
+However, this example is arguably a workflow design error.
+
+If a task generates multiple outputs it is possible to say whether or not at
+least one output is expected - that is down to the internal functioning of the
+task. And if at least one is expected but it doesn't happen, then the task is
+incomplete.
+
+But if `b1` and `b2` above are both success-optional independent tasks, then
+surely two failures at once is a perfectly possible normal outcome, in which
+case it would make sense for the joining dependency to not be triggered.
```
dpmatthews, line 725: It is a possible outcome but it may not be what the user requires. For example, `b1` and `b2` could be tasks that go and retrieve data from sources that may not always be available. `c` can run so long as at least  one of `b1` and `b2` succeed but the workflow can't complete if both fail. This seems a perfectly reasonable scenario to me.

docs/proposal-new-output-syntax.md
dpmatthews: In the latest version you've said: "`foo` means `foo:succeed` UNLESS ..."
That's only true when `foo` is the RHS expresssion, e.g. `bar => foo`.
The "unless" doesn't apply if it's the LHS, e.g.  `foo => bar` - in this case it always means `foo:succeed`.
So, it's a bit awkward to explain.
You've also said "For complete safety, if success is optional use explicit syntax throughout."
If that's the case (and especially if we're going to give validation warnings about this) why not make it mandatory?

I'm not sure I understand the bit about supporting Cylc 7 graphs but we can handle that via the file name (`suite.rc`) if necessary.

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,725 @@
...
+- a platform that isn't always available: `foo:submit-fail?`
+- a task that you expect to fail sometimes: `foo:fail?`
+
+Failure can't be required. Expecting a task to fail every time and treating
+it as incomplete if it succeeds would essentially amount to reversing normal 
+success and failure symantics. If you really have an executable that returns
+non-`0` whilst still doing what you need in the workflow, provide a wrapper
+that correctly returns success status.
+- `foo:fail  # ERROR`
+- `foo:submit-fail  # ERROR`
```
oliver-sanders, line 79: How are we going to provide Cylc 7 back-compat for this?

As expected `:fail` is commonly used, less expected, I found a fair few workflows using `:submit-fail` in OR conditionals.

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,725 @@
...
+```
+A:succeed-any => b
+```
+- trigger b if any member succeeds; but any member that runs is required to
+  succeed
+
+```
+A:succeed-any? => b
+```
+- trigger b if any member succeeds; but it's OK if any or all members fail
```
dpmatthews, line 180: Do you mean that `succeed` is optional for all members of the family?
If so, consider this workflow:
```
a => b1? & b2
b1? | b2 => c
```
Here `b1` may fail but `b2` is expected to succeed. `c` should run as soon as either succeed.
We wouldn't be able to use `B:succeed-any? => c` in this case (because  `succeed` would be both required and optional for `b2`).

The alternative would be to say that `:succeed-any?` doesn't imply anything about the family member outputs (same for `:succeed-all`?).

Not sure what I think about this.

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,725 @@
...
+for all optional output groups.
+
+## Suicide Triggers Still Required at Cylc 8?
+
+Aside from backward compatability we have identified only one remaining problem
+that still requires suicide trigger in Cylc 8: to handle the unwanted extra
+dependencies picked up by a task if it appears on multiple graph branches at
+once (see the Appendix below, and the check-d example above). This will be
+handled cleanly by the dynamic subgraphs planned for Cylc 9; in the meantime
+suicide triggers (and artificial dependencies) provide a decent workaround.
```
dpmatthews, line 403: I may have thought of another?
```
a => b1 & b2? => c
```
We only want `c` to run if both `b1` and `b2` succeed.
We would need to add `b2:fail? => !c` to achieve this.

docs/proposal-new-output-syntax.md
dpmatthews: I think it depends on whether we can infer that the output is optional from the suicide triggers.
I'm not convinced we can do this.
If we can't then I think we should allow this for Cylc 7 workflows (`suite.rc`) and make it an error for Cylc 8 workflows (`flow.cylc`).

docs/proposal-new-output-syntax.md
dpmatthews: I think this example still relies on activating the suicide of `d2` to avoid getting partially satisfied dependencies.

dpmatthews: I remain unconvinced that we can reliably infer optional outputs based on suicide triggers.
Consider this Cylc 8 example:

```
a? => b1 => c1
a:fail? | a:submit-fail? => b2 => c2
c1 | c2 => d
```

At Cylc 7 this might be:
```
a => b1 => c1
a:fail | a:submit-fail => b2 => c2
c1 | c2 => d
d => !a & !b1 & !c1 & !b2 & !c2
```
At Cylc 7 there was nothing to stop users crudely suiciding a whole set of tasks (even if some of them completed successfully) rather than trying to work out some complicated logic to ensure a task only got suicided if necessary. Even if the optimal suicide triggers were used in this case I struggle to see how you would end up with the correct optional outputs.

So, I think we have to action suicide triggers when the pre-requisites are met and define what happens if a task is suicided (see the logic I proposed in #128 which may still be relevant?).
We know there are cases where we need this in any case for some Cylc 8 workflows.

I think we also have to treat Cylc 7 workflows (`suite.rc`) differently from Cylc 8 workflows (`flow.cylc`) for validation purposes.

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,725 @@
...
+This works because Cylc 8 suicide triggers spawn (or update) children just like
+normal triggers. Once all of a task's suicide prerequisites are satisfied it
+is removed immediately. A task that has a single suicide trigger and would not
+otherwise be spawned in Cylc 8 will have a transient presence in the task pool
+because the suicide trigger that spawns it also satisfies its only suicide
+prerequisite.
+
+It doesn't matter if a task still has unsatisfied normal prerequisites when it
+suicides. The other triggers can't spawn it again in the same flow thanks to
+our conditional reflow prevention mechanism.
```
dpmatthews, line 306: That's clever. I hope we retain this behaviour if we make any changes to how suicide triggers are handled.

hjoliver: Not planning to change this at all :+1:

docs/proposal-new-output-syntax.md
hjoliver: Yes, for this scenario it makes sense to have users get rid of the partially-satisfied prerequisite.

docs/proposal-new-output-syntax.md
hjoliver: I'm OK with "failure can't be required" because of the argument that it reverses normal success/fail semantics, which is a variant of the argument that failure is special because by definition it means - or should mean - that the task failed to do what it was supposed to do ... so being "supposed to fail" is perverse.

I think inferring `foo:fail?` in Cylc 7 graphs is quite easy, no need to look at suicide triggers: if both `foo:succeed` and `foo:fail` appear in the graph, they must be optional.  I doubt that anyone is using failure-required semantics (?).

docs/proposal-new-output-syntax.md
hjoliver: > Do you mean that succeed is optional for all members of the family?

Yes. I was thinking that the output syntax on a family would allow us to set **all** member to required or optional, and any exceptions to that would have to be handled by singling out individual members (as we already have to do with family triggers, if individual members have different dependencies).

docs/proposal-new-output-syntax.md
hjoliver: I think that case will work fine. We don't need to use suicide triggers to infer optional outputs because the presence of both `a:succeed` and `a:fail` implies success must be optional.  In addition to that, suicide triggers still work (by the spawn-on-demand logic that you commented on below). But if we infer optional outputs we won't end up with incomplete tasks - which would others indicate a serious error - that eventually get removed by suicide triggers.

hjoliver: So the only difference for Cylc 7 suites is we encounter both `foo:succeed` and `foo:fail` we infer optionality rather than flagging it as an error.

docs/proposal-new-output-syntax.md
hjoliver: OK so you want the workflow to stall if both b1 and b2 fail, even though that is to be expected sometimes, and when it does  occur there would be nothing for `c` to do. To me that seems like a case where it makes perfect sense to shut down cleanly with nothing more to do, rather than stall.

If you want to be alerted to the fact that both b1 and b2 failed, then expecting `c` to be spawned as unsatisfied to cause an eventual stall seems a funny way to do it (Cylc 7 Stockholm syndrome? :grin: ). It would be better to do this:

```
b1? & b2? => c
b1:fail? & b2:fail? => alert
```

hjoliver: Note that in the single task case: 
```
b:x => m
b:y => n
m | n => z
```
... I've been assuming that we can say, based on knowledge of what `b` does, that something has gone wrong if `b` does not complete one output or the other. Either `b` encountered an error but did report failure, or the workflow writer has misunderstood how `b` is supposed to behave.  In which case, having `x | y` as an output completion condition would allow us to flag `b` as incomplete.

If on the other hand not completing either x or y is a possible normal outcome of `b`, then we're in exactly the same position as your two-task example, and it wouldn't be appropriate to use the output completion condition in that case.

hjoliver: (I'll update the text on this)

docs/proposal-new-output-syntax.md
dpmatthews: > any exceptions to that would have to be handled by singling out individual members

You can't do this can you? If the family trigger has made the succeed optional then you can't change it to be required (outputs can't be both required and optional).

docs/proposal-new-output-syntax.md
dpmatthews: > So the only difference for Cylc 7 suites is if we encounter both foo:succeed and foo:fail we infer optionality rather than flagging it as an error.

OK - that's fine

> We don't need to use suicide triggers to infer optional outputs

Later in the proposal you say
"all we need to do is, once again, infer that x, y, and z are optional outputs (the suicide triggers imply this)".
So are you proposing to use suicide triggers to infer optional outputs in other cases?

Are we agreed that:
1) We will treat Cylc 7 workflows differently from Cylc 8 workflows for validation purposes?
2) We will identify Cylc 7 workflows  via the filename (`suite.rc`)?

docs/proposal-new-output-syntax.md
dpmatthews: > OK so you want the workflow to stall if both b1 and b2 fail, even though that is to be expected sometimes, and when it does occur there would be nothing for c to do. To me that seems like a case where it makes perfect sense to shut down cleanly with nothing more to do, rather than stall.

When we have unhandled task failures we stall to allow the user to intervene - the fact that the failure is unhandled doesn't necessarily mean it is unexpected. I view this example in the same way. It's not really important whether the situation (both tasks failing) is expected - what's important is whether the workflow is designed to handle the situation. If it isn't then we should stall to allow intervention.

(This point doesn't really affect the proposal so not a big deal if we view it slightly differently)

docs/proposal-new-output-syntax.md
dpmatthews: "Once all of a task's suicide prerequisites are satisfied it is removed immediately"
Presumably not if it is in the `preparing`, `submitted` or `running` state?
In this case I assume we need to prevent further tasks being spawned (remove it from all flows?) and log a warning (since it doesn't make sense for a task in these states to be suicided).

hjoliver: >  Presumably not if it is in the preparing, submitted or running state?

Yes, we remove it immediately even then.  Even in Cylc 7 we do that actually.  It doesn't cause any problems. You'll get a warning about suiciding an active task.  Task messages will be logged and ignored, from the orphans.

docs/proposal-new-output-syntax.md
hjoliver: > what's important is whether the workflow is designed to handle the situation

Bearing in mind this discussion relates to my proprosed "output completion condition" extension ... in the single task multiple output case, we can *know* that the task is always supposed to generate on output or the other, so if it doesn't generate either one that's a definite error.  But in the two-task case we know that it is quite possible to for both to fail, and I don't see why we should assume that's a stall-worthy problem rather than a possible normal outcome for which the joining task should not be spawned at all.

(But yes, we can't punt this discussion to the future as it won't be in the initial implementation if at all)

docs/proposal-new-output-syntax.md
hjoliver: > Later in the proposal you say
"all we need to do is, once again, infer that x, y, and z are optional outputs (the suicide triggers imply this)".
So are you proposing to use suicide triggers to infer optional outputs in other cases?

Sorry if I wasn't sufficiently clear on this. It's true that the presence of suicide triggers implies there must be optional outputs involved, but in the succeed/fail case we don't need to use them to infer optional outputs. Also, the suicide triggers will still work if they need to.  In the succeed/fail case we can avoid the need though, by inferring optionality.

Branches off of custom outputs might be concurrent or alternate, because we don't know if the outputs are mutually exclusive or not. We could:
- try to infer optional cases from suicide triggers (might be tricky in some cases)
- leave them all as required outputs and let the existing suicide triggers clean up the resulting incomplete tasks

We can start with the second option, and consider trying the first later if it matters.

hjoliver: > Are we agreed that:
> 
>     We will treat Cylc 7 workflows differently from Cylc 8 workflows for validation purposes?

Yes, we can't say use of both `foo:succeed` and `foo:fail` is an error in the Cylc 7 case.

>    We will identify Cylc 7 workflows via the filename (suite.rc)?
 
Presumably that's the easiest thing to do, but I haven't thought about that yet.

docs/proposal-new-output-syntax.md
hjoliver: OK families require more thought. The problem is `succeed-all` etc. are not actually task outputs. They're trigger conditions based on some collective measure of member outputs.

docs/proposal-new-output-syntax.md
hjoliver: >  If that's the case (and especially if we're going to give validation warnings about this) why not make it mandatory?

Yep, now mandatory

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,765 @@
...
+triggers to remove unused branches that are really optional.
+
+In Cylc 8, suicide triggers mostly aren't needed because unused branches don't
+get spawned at all, but they're still there and still if needed.
+
+So if a Cylc 7 workflow is detected (via `suite.rc` vs `flow.cylc` config
+filename):
+- instead of flagging use of both `foo:succeed` and `foo:fail` as an error,
+we can infer that success must be optional for `foo`, to avoid ending up
+with an incomplete task due to the unused output
```
MetRonnie, line 206: Currently users get a warning that the `suite.rc` filename is deprecated, so if they rename their `suite.rc` to `flow.cylc`, they will lose this compatibility. Perhaps we will need a Cylc 7 compatibility mode based on `suite.rc` detection that loudly warns about the graphing changes, and that the changes will need to be made if renaming to `flow.cylc`. Or will it be better to just infer optional success for `flow.cylc` workflows as well, with a warning that this inference is deprecated?

dpmatthews: We're going to have to distinguish between Cylc 7 and 8 workflows for cylc/cylc-flow#2324 as well so we need to treat `suite.rc` and `flow.cylc` files differently. I agree we'll need to review the deprecation message.

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,765 @@
...
+- `foo:fail  # ERROR`
+- `foo:submit-fail  # ERROR`
+
+Finally, optional `start` doesn't really make sense. At best it would be
+equivalent to `foo:submit?`, so use that instead.
+- `foo:start?  # ERROR`
+
+If a task's success is not referenced anywhere in the graph (i.e. only other
+outputs are referenced) then by definition other tasks do not depend on its
+success, and we can therefore safely assume that success is optional. E.g.:
```
dpmatthews, line 64: What about `submit`? I think we should assume `submit` is required unless it is explicitly made optional.

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,765 @@
...
+it as incomplete if it succeeds would essentially amount to reversing normal 
+success and failure symantics. If you really have an executable that returns
+non-`0` whilst still doing what you need in the workflow, provide a wrapper
+that correctly returns success status.
+- `foo:fail  # ERROR`
+- `foo:submit-fail  # ERROR`
+
+Finally, optional `start` doesn't really make sense. At best it would be
+equivalent to `foo:submit?`, so use that instead.
+- `foo:start?  # ERROR`
```
dpmatthews, line 60: Same for `foo:finish?` I assume

dpmatthews: Actually, surely we do need to allow both `start?` and `finish?` if you've used `submit?`.
So maybe any use of `submit?`, `submit-fail?`, `start?` or `finish?` implies that all outputs from that task must be optional?

docs/proposal-new-output-syntax.md
hjoliver: Yeah. OK, changing this for `succeed` too, to make for fewer exceptions.  i.e. if `foo:succeed` or `foo:fail` not mentioned anywhere in the graph (only `foo:x` perhaps) assume that success is required.

docs/proposal-new-output-syntax.md
hjoliver: Noted this as unresolved in the proposal, with another possible option.

docs/proposal-new-output-syntax.md
hjoliver: `:finish` is a pseudo-output, like a family trigger.  I think `foo:finish?` doesn't make sense - what does it mean for "finishing" to be optional when it encompasses both success and failure?

hjoliver: `:submit?` (and `submit:fail?`) don't imply that subsequent task-execution outputs have to be optional.  You might want to handle job submission failure, but if the job submits OK you want success to be required (e.g.).

hjoliver: Note tasks can only be checked for completion when they have, erm, "come to a stop" ... which means either job submission failed, or execution finished.

hjoliver: `:start?` seems pointless at best.  A task can't succeed or fail (or complete any other outputs) if it never started.

dpmatthews: > `:submit?` (and `submit:fail?`) don't imply that subsequent task-execution outputs have to be optional. You might want to handle job submission failure, but if the job submits OK you want success to be required (e.g.).

Ah, of course - I hadn't thought that through. That makes sense and I see you've updated the proposal to make this clear - thanks. I agree `:finish?` makes no sense. I think the same is true of `:start?`

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,783 @@
...
+- for other outputs, e.g. `foo:x` and `foo:y` we can't know if they're optional
+  or not, so assume they are required and let the existing Cylc 7 suicide
+triggers clean up any resulting incomplete tasks 
+
+### How to detect Cylc 7 workflows?
+
+- detect `suite.rc` vs `flow.cylc` config filename?
+  - users might change the filename without upgrading the syntax?
+- require users to add an explicit marker comment in the file?
+- other?
```
dpmatthews, line 228: If users change the filename without upgrading they will get validation errors for any problems in the graph so will soon realise their mistake. I don't really see why we need to allow users to be able to create `flow.cylc` files with Cylc 7 graphs.

hjoliver: Ok I’m happy with that 👍🏻

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,783 @@
...
+
+## Suicide Triggers Still Required at Cylc 8
+
+Suicide triggers will probably be made entirely obsolete by dynamic subgraphs
+at Cylc 9. In the meantime we need to keep them for the following reasons:
+
+- primarily, backward compatibility for Cylc 7 graphs
+
+- secondarily, to allow users to housekeep partially satisfied prerequisites
+  that can occur due to partial dependence on optional outputs
```
dpmatthews, line 451: Or due to submit failures, e.g.
```
a & b => c
b:submit-fail? => !c
```

hjoliver: :+1: updated

docs/proposal-new-output-syntax.md
hjoliver: (We've gone with `foo:fail` allowed, after discussion on Element)

docs/proposal-new-output-syntax.md
hjoliver: (Family proposal updated after discussion on Element chat).

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,769 @@
...
+A:x-all => b  # member x required
+A:x-any => b  # member x optional
+```
+
+To mess with the defaults you can refer to member outputs collectively or
+individually, anywhere in the graph, e.g.:
+```
+# Family outputs (note these cannot be family triggers)
+A?  # member success optional
+A:x?  # member x optional
```
dpmatthews, line 185: Is it worth supporting this special syntax when you can use , e.g, `A:succeed-any` for the same purpose?
Note that you'd have to use `A:fail-all` to set the default to failure required.

Also, we need to define what happens when 2 family triggers imply different defaults:
```
A => b
A:fail-all => c
```
In the case of mixed defaults I think we should interpret that to mean the default is optional in all cases.

docs/proposal-new-output-syntax.md
```diff
@@ -0,0 +1,769 @@
...
+A:fail-any => b  # member success/failure optional
+
+A:finish-all => b  # member success optional
+A:finish-any => b  # member success optional
+
+A:start-all => b  # N/A
+A:start-any => b  # N/A
+
+A:submit-all => b  # member submit success required
+A:submit-any => b  # member success optional
```
dpmatthews, line 171: "success" should be "submit/submit-fail"

docs/proposal-new-output-syntax.md
hjoliver: We don't actually support `A => b` BTW, have to be explicit with family triggers.

But I agree on "mixed defaults" - forgot to add that to the proposal sorry... for next update

docs/proposal-new-output-syntax.md
hjoliver: :+1:  Oops, cut-n-paste fail.

docs/proposal-new-output-syntax.md
dpmatthews: What about the special syntax? (no big deal)

hjoliver: I was thinking this "special syntax" would be more straightforward and intuitive than using the family pseudo-output trigger syntax just to set output type via the defaults (which vary according to family trigger).  However, not entirely sure about that so I've done the implementation without the special syntax, at least initially.

hjoliver: text updated

dpmatthews (approved):
All good now I think!

hjoliver:
Merging this now - only one official approval, but agreed offline and already implemented.

PR #131, opened 2021-08-04
add link to open questions

oliver-sanders:
The idea being that the "question" label should be a flag for the cylc/core team to get involved.

We should shoehorn in a review of these open questions in each team meeting to try and avoid letting them build up too long.

Might take a little while to crack down on the number.


hjoliver (approved):
:+1:

PR #140, opened 2021-12-10
remove my very dated notes on the UI 

wxtim:
Use repo readme files as they are more up to date


datamel (commented):
I wonder if this would be better updated than removed. The reason for this is that repos link to this section for how to set up developer environment for contributing. 
Looking at e.g. the README on cylc-ui, it assumes you have your environment set up with cylc-flow and yarn etc all ready. 
I think it would be sensible to have a definitive guide on how to set up the whole environment, since we would not want to frustrate potential contributors. I myself have had a look at these recently in order to try and fix probs with my env so nice to have a working reference.
As always, happy to be over-ruled!

wxtim:
> I wonder if this would be better updated than removed. The reason for this is that repos link to this section for how to set up developer environment for contributing. Looking at e.g. the README on cylc-ui, it assumes you have your environment set up with cylc-flow and yarn etc all ready. I think it would be sensible to have a definitive guide on how to set up the whole environment, since we would not want to frustrate potential contributors. I myself have had a look at these recently in order to try and fix probs with my env so nice to have a working reference. As always, happy to be over-ruled!

Quite happy to try and keep this up-to-date, but I need to be better at _actually_ doing it.

PR #151, opened 2022-04-05
Add project issue summary

dpmatthews:
I think this provides a useful summary of the project which we can refer to in the VCs.

Note that it will need updating as we close and open milestones but I'm happy to do this.


wxtim:
I like it - Perhaps we should put up an issue to review it in 3 months, deleting if it's been unused/unmaintained.

wxtim (approved):

hjoliver (approved):
Good idea (so long as we keep it up to date!).  It'll simplify the cut-paste-modify routine a bit for meeting agenda.

PR #154, opened 2022-05-19
update release template

oliver-sanders:


hjoliver (approved):
Thanks @oliver-sanders, that's clearer :+1:

PR #167, opened 2023-02-03
add status page

oliver-sanders:
* Add a JSON file mapping component branches to meta releases.
  This will save us having to maintain this list for each of the components individually.
* Add a status page generated from this to give better visibility of failures.

This can be used to automatically pick compatible Cylc versions for installation for CI in the Cylc component repositories via release-actions - https://github.com/cylc/release-actions/pull/46

Screenshot from 2023-02-03 13-24-10

### DIFF
--- a/.github/wokflows/status.yml
+++ b/.github/wokflows/status.yml
@@ -37,10 +37,11 @@ jobs:
     timeout-minutes: 10
 
     steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
       - name: configure python
         uses: actions/setup-python@v4
-        with:
-          python-version: '3.11'
 
       - name: install
         run: pip install jinja2
--- a/.github/wokflows/status.yml
+++ b/.github/wokflows/status.yml
@@ -1,0 +1,61 @@
+# THIS FILE IS PART OF THE CYLC WORKFLOW ENGINE.
+# Copyright (C) NIWA & British Crown (Met Office) & Contributors.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Deploy a new version of the documentation or rebuild an existing one.
+
+# Note: This action deletes its previous commits so the nightly build history
+# is not preserved and does not require housekeeping.
+
+name: Status
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '35 0 * * *'
+  push:
+    branches:
+      - 'master'
+    paths:
+      - 'docs/status/*'
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+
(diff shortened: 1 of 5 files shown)


oliver-sanders:
For some reason this action doesn't show up on my fork when pushed to master - https://github.com/oliver-sanders/cylc-admin/actions

Not sure why, I've probably done something badly wrong, not sure how to test the action without it.

MetRonnie:
Do you need to enable it in the repo settings?

oliver-sanders:
I think it's enabled?

Screenshot from 2023-02-03 15-18-40

MetRonnie:
Only thing I can think of is to push a dummy commit to master to try and get it to trigger the Actions workflow for the first time

oliver-sanders:
Good idea, sadly didn't do the trick.

oliver-sanders:
It's not a complex action and the script can be tested manually so can always fix in situ if needed.

.github/wokflows/status.yml
```diff
@@ -0,0 +1,61 @@
...
+      - 'master'
+    paths:
+      - 'docs/status/*'
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+
+    steps:
```
MetRonnie, line 39: I think it's missing the checkout step?

.github/wokflows/status.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: configure python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
```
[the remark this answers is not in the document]
oliver-sanders: Not really I just thought that controlling this explicitly might result in fewer surprises.

Happy to drop though.

.github/wokflows/status.yml
oliver-sanders:
```suggestion
    steps:
      - name: Checkout
        uses: actions/checkout@v3

```

.github/wokflows/status.yml
```diff
@@ -0,0 +1,62 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Deploy a new version of the documentation or rebuild an existing one.
+
+# Note: This action deletes its previous commits so the nightly build history
+# is not preserved and does not require housekeeping.
```
MetRonnie, line 20: Doesn't look like it is doing this to me

oliver-sanders, line 20:
```suggestion
```

MetRonnie:
>```
># Note: This action deletes its previous commits so the nightly build history
># is not preserved and does not require housekeeping.
>```
> Doesn't look like it is doing this to me

I see you've now removed that comment, but isn't self-housekeeping desirable?

oliver-sanders:
> isn't self-housekeeping desirable?

That would require force pushing.

That comment was accidentally copied from a cylc-doc action where each "push" contains a complete build of the docs which is substantial. Each of these commits is a couple of lines of JSON and Markdown so much less concerning.

oliver-sanders:
> IMO having nightly commits piling up would be a bit messy,

We shouldn't actually get "nightly" commits because the status page should only change when the branches get updated. Otherwise the page should build exactly the same every time so there will be nothing to commit, so nothing will get pushed up. We should only get one commit each time we change the branches, typically once per minor release of cylc-flow which will be a manageable number.

I think workflow should probably skip the "push changes" part if there's nothing to commit incase the commit or push return non-zero because of this?

Note, the cron schedule isn't actually necessary here, the only reason I added it was to have some indication if the action breaks.

hjoliver (approved):
Very nice @oliver-sanders

PR #171, opened 2023-04-04
Fix status page

MetRonnie:
- Try using HTML table to see if that fixes render on GH Pages
- Fix cylc-flow listed workflows

### DIFF
--- a/docs/status/branches.json
+++ b/docs/status/branches.json
@@ -16,7 +16,7 @@
         }
     },
     "branch_actions": {
-        "cylc/cylc-flow": ["test_fast", "test_functional", "build", "test_conda-build"],
+        "cylc/cylc-flow": ["test_fast", "test_functional", "bash", "build", "test_manylinux", "test_conda-build", "test_tutorial_workflow"],
         "cylc/cylc-rose": ["tests"],
         "cylc/cylc-uiserver": ["test", "build"],
         "cylc/cylc-doc": ["test"],
--- a/docs/status/template.md.jinja2
+++ b/docs/status/template.md.jinja2
@@ -1,14 +1,5 @@
-{#
-
-This template/markdown file is auto-generated, please do not edit it directly.
-
-Please edit `branches.json` and raise a pull request, this fille will be
-automatically regenerated on merge.
-
-#}
-
-{% macro badge(repo, branch, action) -%}
-[![{{ action }}](https://github.com/{{ repo }}/actions/workflows/{{ action }}.yml/badge.svg?branch={{ branch }})](https://github.com/{{ repo }}/actions/workflows/{{ action }}.yml)
+{% macro badge(repo, branch, workflow) -%}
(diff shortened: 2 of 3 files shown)


docs/status/template.md.jinja2
```diff
@@ -1,14 +1,5 @@
-{#
-
-This template/markdown file is auto-generated, please do not edit it directly.
-
-Please edit `branches.json` and raise a pull request, this fille will be
-automatically regenerated on merge.
-
-#}
```
MetRonnie, line 8: I've moved this comment to the auto-generate file

docs/status/branches.json
```diff
@@ -16,7 +16,7 @@
         }
     },
     "branch_actions": {
-        "cylc/cylc-flow": ["test_fast", "test_functional", "build", "test_conda-build"],
+        "cylc/cylc-flow": ["test_fast", "test_functional", "bash", "test_manylinux", "test_conda-build", "test_tutorial_workflow"],
```
oliver-sanders, line 19: "build" removed purposefully?

docs/status/template.md.jinja2
oliver-sanders: Why? Not much point putting it there!

docs/status/template.md.jinja2
```diff
@@ -1,14 +1,5 @@
...
-
-{% macro badge(repo, branch, action) -%}
-[![{{ action }}](https://github.com/{{ repo }}/actions/workflows/{{ action }}.yml/badge.svg?branch={{ branch }})](https://github.com/{{ repo }}/actions/workflows/{{ action }}.yml)
+{% macro badge(repo, branch, workflow) -%}
```
oliver-sanders, line 1: Why rename every instance of action to workflow?!?!?!

MetRonnie: Because they're workflows not actions

docs/status/branches.json
MetRonnie: There isn't a `build.yml` workflow

docs/status/template.md.jinja2
MetRonnie: But isn't that where it was meant to go? The template isn't the one that's autogenerated

docs/status/template.md.jinja2
oliver-sanders: Yes and no. With this change there's now a confusing mix of terminology within the same file which is much worse.

docs/status/branches.json
oliver-sanders: MB, the workflow was added on a PR which didn't get merged so it showed in GH but wasn't on master. I've cherry-picked that one onto master so it's there again now.

docs/status/template.md.jinja2
oliver-sanders: Derp.

docs/status/branches.json
MetRonnie: Yeah I saw, added back in after rebase

docs/status/template.md.jinja2
MetRonnie: Is it too late to rename `nightly_actions` and `branch_actions`?

oliver-sanders: If you really want to go through all the downstream workflows, but there are much more worthwhile uses of our time.

PR #175, opened 2023-07-27
minor suggestion

wxtim:
If the default is to disable event handlers in skip mode it makes more sense to me for the setting the user adds to be called `enable....`.

### DIFF
--- a/docs/proposal-skip-mode.md
+++ b/docs/proposal-skip-mode.md
@@ -43,7 +43,7 @@ apply only to the pool and Cylc 7 style reset is not permitted.
    The valid configurations would be:
    * `outputs` - Define the outputs to be generated when this task runs
      in skip mode. By default, all required outputs will be generated.
-   * `disable task event handlers` - Disable the event handlers which would
+   * `enable task event handlers` - Enable the event handlers which would
      normally be called on task lifecycle events. By default event handers
      will be turned off.
 


docs/proposal-skip-mode.md
```diff
@@ -43,7 +43,7 @@ apply only to the pool and Cylc 7 style reset is not permitted.
    The valid configurations would be:
    * `outputs` - Define the outputs to be generated when this task runs
      in skip mode. By default, all required outputs will be generated.
-   * `disable task event handlers` - Disable the event handlers which would
+   * `enable task event handlers` - Enable the event handlers which would
      normally be called on task lifecycle events. By default event handers
      will be turned off.
```
oliver-sanders, line 48: I kinda get the point about negative configurations, however, Cylc (and most tools) make use of negative configurations e.g:

```
cylc play --no-detach
cylc vip --no-timestamps
cylc clean --no-scan
cylc install --no-run-name
```

This also matches the existing `[runtime][<namespace>][simulation]disable task event handlers` and makes it clear that these handlers are enabled by default (because you have to turn on the disable setting to turn them on).

wxtim: > cylc play --no-detach &c

All the examples are negatives where the negative sets a `variable = False` _and_ False is not default. 

> makes it clear that these handlers are enabled by default

They aren't enabled by default in simulation mode and won't be in skip mode either.

>  you have to turn on the disable setting to turn them on

Does that not strike you as as odd? (Yes I understand that you are saying `disable=False`)

I don't really care about this. I certainly can't be bothered with the logic required to have a back compatibility for sim mode section (i.e. keep both settings and work out the truth table of what happens if the user chooses one or both settings(`WARNING - you set disable task handlers = False and enable task handlers = False what were you thinking?`)).

I'm happy to let the matter drop for the sake of consistency and accept that the time to have challenged this was pre-8.0 when we were making changes.

PR #181, opened 2023-10-09
Modify the optional outputs extension proposal

dpmatthews:
A replacement for the optional outputs extension agreed in  #173.

### DIFF
--- a/docs/proposal-optional-output-extension.md
+++ b/docs/proposal-optional-output-extension.md
@@ -2,13 +2,11 @@
 
 A proposal which extends the
 [optional outputs model](https://cylc.github.io/cylc-admin/proposal-new-output-syntax.html)
-to solve existing limitations and support the expired state by providing more
-information to Cylc about the user's intentions for graph branching.
+to solve existing limitations and support the expired state.
 
 Supersedes:
-* [cylc-admin/task expire proposal](https://github.com/cylc/cylc-admin/blob/fc564ddb26c1476dd051b059ca9a31829a20bf30/docs/proposal-task-expire.md)
+* [cylc-admin/previous optional output extension proposal](https://github.com/cylc/cylc-admin/blob/4c5bf7f50125472a1c022ad9108ac7fd2a7ec8bf/docs/proposal-optional-output-extension.md)
 * [cylc-flow#5423](https://github.com/cylc/cylc-flow/issues/5423)
-* [cylc-flow#3294](https://github.com/cylc/cylc-flow/issues/3294)
 
 
 ## Background: Task Expiry
@@ -22,17 +20,8 @@ There are currently three problems with expiry (as of cylc-flow 8.1.4):
 
 2. **Expiry triggers don't fit in with [optional outputs](https://github.com/cylc/cylc-flow/issues/5361)**
 
-   Under the existing logic, if expiry is optional, then success must be too which means
-   that failure would go uncaught making expiry unsafe.
-
-   There are fundamentally two solutions to this problem.
-
-   1. Move expiry to a separate model.
-      * The existing
-        [proposal-task-expire](https://cylc.github.io/cylc-admin/proposal-task-expire.html).
-        follows this approach.
-   2. Modify optional output logic to allow expiry to fit.
-      * The approach this proposal follows.
+   Under the existing logic, task expiry cannot cause incomplete tasks. This makes expiry unsafe.
+   This proposal addresses this.
 
 3. **Expiry events can be detected late, or even missed due to the task pool implementation.**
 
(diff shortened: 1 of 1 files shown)


docs/proposal-optional-output-extension.md
```diff
@@ -41,52 +30,33 @@ There are currently three problems with expiry (as of cylc-flow 8.1.4):
...
-and "problem" outcomes.
-
-
 ## Proposal
 
 1. Expiry should remain implemented as a task output.
 
-   If a task is clock-expired OR if `expired` output is set manually, then the
-   task status should change to expired as per
-   [cylc set proposal](https://cylc.github.io/cylc-admin/proposal-new-output-syntax.html).
```
oliver-sanders, line 73: This statement hasn't been revoked, just removed from the proposal as orthogonal.

oliver-sanders (approved):
This proposal handles the expire problem by subtly redefining the meaning of "required" to cover execution only, i.e. if this task runs it must succeed, but it might not run; rather than handing both at the same time, i.e. this task might not succeed because it might not run.

Doing this avoids the three-way problem of succeeded-failed-expired by reducing completion states to two-way switches which we can handle more easily i.e:

* * succeeded
   * or failed
* * submitted
   * or submit-failed
* * expired
   * or not expired

It's a detriment to graph branching uses over the previous as it drops the default output validation, however, avoiding an extra `??` graph syntax is probably a more valuable outcome here.

Completion expressions remain as a solution to output validation, but are no longer core to the implementation as they only cover execution so cannot serve as the single source of truth for task completion. Additional branched logic is required to handle the pre-conditions to task execution (submitted / not-expired). Given that the completion expression work has already been done, we might as well use it, but it's not strictly required any more.

The expire case is handled safely by requiring it to be explicitly handled in the graph brining it to parity with submitted/submit-failed which is fundamentally similar.

I think this is a better response to the "irk" with the previous proposal than trying to shim extra syntax into the model to avoid complicating this further.

docs/proposal-optional-output-extension.md
```diff
@@ -292,55 +217,44 @@ a:z? => z
...
-completion(a) = succeeded or expired
+a => x
+a:expired?
 ```
 
-Early halt in the event of expiry now works the same as with optional outputs.
-Early halt in this example is now explicit in that it follows an optional output, making it clear
-that the `x` branch might not happen and that expiry must be handled if early halt is not desirable.
+`a:expired?` has to be included in the graph to avoid validation failure.
+This makes it clear that the halt is intentional.
```
MetRonnie, line 257: What exactly is meant by "halt"?

MetRonnie (commented):
I like this. It seems logically consistent and minimal change

hjoliver (approved):
All good, I'm happy with this.

For optional outputs, it doesn't go quite as far in protecting users from themselves as the originally-agreed (sorry) proposal, but I think it's simpler and nothing sticks out as inconsistent (like an optional output not really being optional by default).

For expiry, it's like my original proposal but with the additional safety net of forcing explicit use of `:expire?` in the graph, even for intended "early halt", which is all good.

hjoliver:
> For expiry, it's like my original proposal but with the additional safety net ...

It doesn't really matter now, but contrary to what it says (but not added in this PR) under the section "Preference To The Expire Proposal" my original proposal didn't really "recommend converting expiry from an output to a task attribute".

The main points of it were:
- expiry should not cause a stall (i.e. it should not cause incomplete tasks)
- expiry must be optional
- optional expire does not imply optional succeed

On the last point, it said:

> Required outputs are always conditional on the owner task actually running, and an expired task does not run at all. So :succeed (and other outputs) can still be required even if expiration is optional. Required success really means, if the task runs, its success is required.

Unfortunately I neglected to flag `:submit-fail` as being similar to `:expire` in this respect, doh.

And also:

> NOTE this is moot if [**IF**] we decide that :expire should not be considered a task output - see below.

i.e., while I argued at the bottom that expired should really be an attribute not a state/output, but that wasn't supposed to be a fundamental part of the proposal.

Anyway, I'm happier now, because expiry is more like originally intended AND safer!

hjoliver:
Actually, before we merge this:

One thing not explicitly addressed is forced expire - i.e., manually tell the scheduler not to bother running this task.

To protect against early halt on overzealous use of forced expiry, I presume we should handle it like this:
- if `:expire` is not optional, create an incomplete task
- otherwise, complete the task (as expired) and remove it

dpmatthews:
> Forced expiry is not explicitly addressed here - i.e., manually telling the scheduler not to bother running a task.

Expiry is expiry - it makes no difference whether it is set via `cylc set`. So, the normal condition applies: a task in incomplete If the task expired and the` :expire` output was not optional.

If you want to not run a task, don't use expiry unless you've configured the workflow to handle it. You can use `cylc set` or skip mode or possibly `cylc remove` depending on what you're trying to achieve.

hjoliver:
In that case, merging 🎉

PR #183, opened 2023-10-17
proposal set: fix small error

oliver-sanders:
Sorry, spotted a small error in the `cylc set` proposal.

`cylc set --pre=all` is NOT equivalent to trigger as outlined in "[7. Spawning parentless tasks](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-cylc-set.md#7-spawning-parentless-tasks)". It's more equivalent to Cylc 7's insert.

### DIFF
--- a/docs/proposal-cylc-set.md
+++ b/docs/proposal-cylc-set.md
@@ -72,7 +72,8 @@ By default this sets all required outputs for the given task(s).
    --pre=PRE: Set a prerequisite to satisfied e.g. `foo:succeeded`
               (multiple use allowed, may be comma separated).
 
-   --pre=all: Set all prerequisites to satisfied. Equivalent to trigger.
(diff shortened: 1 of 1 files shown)


hjoliver (approved):

hjoliver:
The way I put it in the `cylc set` PR is:
```
 Setting prerequisites contributes to a task's readiness to run and promotes it
to the active window where any clock and xtriggers will become active.
```
Which in fact applies to setting any or all prerequisites.

 I'll go back and compare my wording (in the PR) to yours (in the proposal) to see what's clearest ...

hjoliver:
> It's more equivalent to Cylc 7's insert.

Except unlike insert, in Cylc 8 `--pre=all` IS equivalent to trigger IF there are no non-prerequisite constraints on running.

hjoliver:
So I'll merge this, but the CLI help might still require a bit extra to fully explain.

PR #186, opened 2023-10-24
fix status page

oliver-sanders:
Workflow failed because we need to explicitly provide GITHUB_TOKEN as an environment variable.


wxtim:
Have you run this to check?

oliver-sanders:
No, can't without setting it up to deploy to GH pages.

wxtim (approved):

PR #189, opened 2024-03-15
amendment to proposal amendment

oliver-sanders:
### Summary

* Redefine task "incompletion" to avoid problem edge cases.
* Permit pre-conditions in completion expressions (no longer any reason not to, this makes things clearer for users)
* Remove redundant safety check that was only required for the previous proposal

### *In*completion condition bugs

Closer inspection of the condition for task "incompletion" revealed some logical flaws, e.g:

#### 1. Task cannot be completed by the expired output.

If a task is expired (and expiry is permitted) then it is complete, but subsequently setting a finished output would make it incomplete by satisfying the "it finished executing" condition.

Similarly, if the task succeeds, fails or submit-fails, then it will not be possible to subsequently complete it by setting the expired output.

#### 2. Tasks with no required outputs

It is possible for a task to have no required outputs, e.g:

```
a:submitted?
a:succeeded?
a:expired?
```

In this case the completion condition is `True`. By the previous logic the task is complete before it has even been submitted.

#### Underlying issue

These flaws weren't intentional, they are just bugs. The issue here is the defining *in*completion which results in a lot of double negatives.

For the same reason we cannot use `not` in completion conditions (task outputs accumulate over multiple executions / interventions), we cannot define task completion in terms of negatives.

The solution is to define task *completion* in terms of positives. This is also clearer and easier to implement.

### Implementation detail: Exposing the completion condition

A bonus of the new definition of task completion is that we can define the default completion condition for a task in a single flat expression e.g:

```
( succeeded and x) or expired
```

(As opposed to the previous branched code logic)

This condition can then be exposed to the user clarifying how the optional outputs they define in the graph impact the completion of the task.

E.G:

```
a?
a:failed?
```

```console
$ cylc config -i '[runtime][a]completion'
completion = succeeded or failed
```

### Behaviour change

There has been one functional change in this amendment. Consider:

```
a?
a:failed?
a:x
```

Following the previous logic the task `a` could have been completed in one of two ways:

1. `succeeded` and `x`
2. `failed` and `x`

So the completion condition, effectively reduced to:

```
(succeeded and x) or (failed and x)
```

This isn't especially helpful as we probably wouldn't have expected `x` to be generated in the failure scenario.

I'm not sure whether this is a bug in the condition or purposeful. This amendment changes:

> All required outputs must be completed if the task **runs**. The task must succeed unless stated otherwise.

To:

> All required outputs must be completed if the task **succeeds**. The task must succeed unless stated otherwise.

Changing the default completion expression for the above example to:

```
(succeeded and x) or failed
```

Which is a tad more practical.

### DIFF
--- a/docs/proposal-optional-output-extension.md
+++ b/docs/proposal-optional-output-extension.md
@@ -41,16 +41,33 @@ There are currently three problems with expiry (as of cylc-flow 8.1.4):
    executed and that might not happen if expiry or submission are optional (or if the pre-requisites
    for `a` are not met).
 
-3. A task is incomplete if the completion condition is not satisfied and:
+3. A task is complete if the completion condition is satisfied.
 
-   * It finished executing
+   The completion condition can be defined by the user, otherwise the
+   expression is automatically generated according to the rules:
 
-   * OR Job submission failed and the `:submit` output was not optional
+   1. All required outputs (referenced in the graph) must be satisfied.
+   2. If success is optional, then the task is complete if it fails.
+   3. If submission is optional, then the task is complete if it submit-fails.
+   4. If expiry is optional, then the task is complete if it expires.
 
-   * OR The task expired and the `:expire` output was not optional
-     * This is a new condition to handle expiry
+   I.E:
+
+   ```python
+   is_complete = (
+     all(output for output in outputs if not is_optional(output))
+     or (failed if is_optional(succeeded))
+     or (submit_failed if is_optional(submitted))
+     or (expired if is_optional(expired))
+   )
+   ```
+
+   Note, a task is *in*complete, if the completion condition is not satisfied
+   but the task has a final task status (note status not output). This is the
+   same as presently implemented (8.2.x)
 
 4. `:expire` cannot be required.
+
    Similarly `:submit-fail` cannot be required.
    
    * We currently allow `:submit-fail` to be required but this doesn't really make sense.
@@ -69,6 +86,8 @@ There are currently three problems with expiry (as of cylc-flow 8.1.4):
    # OK
    completion = succeeded or failed
    completion = succeeded and (x or y or z)
+   completion = (succeeded and x) or (failed and y)
+   completion = (succeeded and (x or y or z)) or failed or expired
 
    # ERROR
    completion = not failed
@@ -78,8 +97,8 @@ There are currently three problems with expiry (as of cylc-flow 8.1.4):
 
    Expression to be validated (i.e. parsed) at validate time.
 
-   If the completion expression is defined, any outputs which are optional in the expression
-   should be marked as optional in the graph if used in the graph:
+   If the completion expression is user-defined, then it must be logically
+   consistent with the graph.
 
    ```
    completion(a) = succeeded and (x or y or z)
@@ -114,40 +133,36 @@ There are currently three problems with expiry (as of cylc-flow 8.1.4):
    }
    ```
 
-   Note: the completion condition only covers outputs generated when a task is executed - it cannot
-   include `:expired`, `:submit` or `:submit-fail`. It also cannot include `started` (this has to
-   happen if a task executes) or `finished` (this isn't an output).
+   The completion expression cannot include `finished` (this isn't an output).
 
-   The default condition is complete all required outputs.
+6. Expiry is only optional if it is explicitly referenced in the graph (e.g.
+   `a:expired? => x`) or permitted in the completion expression
+   (e.g. `succeeded or expired`).
 
-6. Expiry is only optional if it is explicitly referenced in the graph e.g. `a:expired? => x`.
-
-   If a task is configured to clock-expire it must be made optional via the graph - otherwise this
-   will cause a validation failure. This is designed to reduce the risk of unintended early shutdown.
-
-   For this example:
+   E.G. in these examples:
 
    ```
-   clock-expire = a
-   a => b
-   a:expired?  # This does nothing but is sufficient to pass validation
-   b:submit-fail? => c
+   a => z
+
+   b => z
+   b:expired?
+
+   c
+   # completion = succeeded or expired
    ```
 
-   Task "a" is completed in one of 2 ways:
+   If the `expired` output is set on both tasks `a` and `b`:
 
-   * It expires
-   * It executes and succeeds
+   * `a` will be incomplete.
+   * `b` will be complete (permitted in graph).
+   * `c` will be complete (permitted in completion expression).
 
-   If task "a" succeeds then task "b" must complete which can happen in one of 2 ways:
-
-   * It fails to submit
-   * It executes and succeeds
-
-   So, the expression `a => b` in the graph can be interpreted as:
-   * If "a" executes it must succeed
-   * If "a" succeeds then "b" must complete
-   * If "b" executes it must succeed
+   If a task is configured to clock-expire, but expiry is not handled in the
+   graph or permitted in the completion expression, a warning should be raised
+   informing the user that the workflow may stall. The warning should
+   recommend that they either handle the expired output in the graph if no
+   completion expression is defined, or permit it in the completion expression
+   if one is defined.
 
 7. If neither `:succeed` nor `:fail` are specified for a task in the graph then `:succeed` should
    be required.
@@ -253,7 +268,7 @@ a => x
 a:expired?
 ```
 
-`a:expired?` has to be included in the graph to avoid validation failure.
+`a:expired?` has to be included in the graph to avoid stall.
 This makes it clear that the halt is intentional.
 
 ### Output Groups


docs/proposal-optional-output-extension.md
```diff
@@ -114,40 +133,36 @@ There are currently three problems with expiry (as of cylc-flow 8.1.4):
    }
    ```
 
-   Note: the completion condition only covers outputs generated when a task is executed - it cannot
-   include `:expired`, `:submit` or `:submit-fail`. It also cannot include `started` (this has to
-   happen if a task executes) or `finished` (this isn't an output).
```
oliver-sanders, line 119: This restriction is no longer necessary with the above change.

We could decide to preserve this restriction, however, it will be clearer for users if we broaden the scope of the completion condition to include pre-conditions as otherwise we will have to automatically add the pre-condition logic internally.

oliver-sanders:
#### Playground:

I have an implementation of the proposed logic on [this branch](https://github.com/cylc/cylc-flow/compare/master...oliver-sanders:cylc-flow:optional-outputs-extension?expand=1), it's not ready yet of course, but you can use it to see how the default completion condition is generated, e.g:

```ini
[scheduling]
    [[graph]]
        R1 = """
            a? & a:x
            a:failed?

            b? & b:x?
            b:failed?

            c? & c:x
            c:failed? & c:y

            d?
            d:submit?
            d:expire?
        """

[runtime]
    [[root]]
        [[[outputs]]]
            x = xxx
            y = yyy
            z = zzz

    [[a]]
    [[b]]
        # user defined completion expression
        # (this can be anything so long as it's consistent
        #  with the graph)
        completion = (succeeded and x) or (failed and y)
    [[c]]
    [[d]]
```

```console
$ cylc config . -i '[runtime]'
[[root]]
    [[[outputs]]]
        x = xxx
        y = yyy
        z = zzz
[[a]]
    completion = (x and succeeded) or failed
    [[[outputs]]]
        x = xxx
        y = yyy
        z = zzz
[[b]]
    completion = (succeeded and x) or (failed and y)
    [[[outputs]]]
        x = xxx
        y = yyy
        z = zzz
[[c]]
    completion = ((x and y) and succeeded) or failed
    [[[outputs]]]
        x = xxx
        y = yyy
        z = zzz
[[d]]
    completion = succeeded or failed or submit_failed or expired
    [[[outputs]]]
        x = xxx
        y = yyy
        z = zzz
```

docs/proposal-optional-output-extension.md
```diff
@@ -41,16 +41,33 @@ There are currently three problems with expiry (as of cylc-flow 8.1.4):
...
 
-   * It finished executing
+   The completion condition can be defined by the user, otherwise the
+   expression is automatically generated according to the rules:
 
-   * OR Job submission failed and the `:submit` output was not optional
+   1. All required outputs (referenced in the graph) must be satisfied.
+   2. If success is optional, then the task is complete if it fails.
+   3. If submission is optional, then the task is complete if it submit-fails.
+   4. If expiry is optional, then the task is complete if it expires.
```
hjoliver, line 52:
```suggestion
   1. If all required outputs (referenced in the graph) are satisfied, then the task is complete if it is finished.
   2. If success is optional, then the task is complete if it fails.
   3. If submission is optional, then the task is complete if it submit-fails.
   4. If expiry is optional, then the task is complete if it expires.
```

hjoliver: Also, we need to decide whether outputs get "completed" or "satisfied". I have always said the former (only prerequisites get satisfied). However I suppose we could say that outputs get satisfied, in order to avoid any confusion with task completion.

oliver-sanders: > These points need ORs

Can add `or`s, it doesn't change the meaning.

> and for 1., the task must be finished, right?

No, we determine whether a task is "finished" from the task state not the task outputs. E.G. a retrying task may have the "failed" task output but be in the "waiting" state. Attempting to build the "finished" check into the completion expression results in bugs.

The completion expression tells you whether a task is complete, a task is *in*complete if this expression evaluates `False` AND the task is in a final *state*.

oliver-sanders:
```suggestion
   1. All required outputs (referenced in the graph) must be satisfied.
   2. Or, if success is optional, then the task is complete if it fails.
   3. Or, if submission is optional, then the task is complete if it submit-fails.
   4. Or, if expiry is optional, then the task is complete if it expires.
```

hjoliver: > No, we determine whether a task is "finished" from the task state not the task outputs...

Yes but no but. By "finished" I did mean a final state, not outputs. 

I missed this critical change a bit higher up, sorry, which partly explains your response:
```diff
- 3. A task is incomplete if the completion condition is not satisfied and:
+ 3. A task is complete if the completion condition is satisfied.
```

This is change in terminology is not ideal, at least not without other attendant changes. 

Currently:
- *a task is complete **if** its outputs are complete **and** it is finished
- *a task is incomplete **if** its outputs are not complete **and** it is finished

[N.B., by "its outputs are complete" I mean either "its required outputs are complete" (currently) or "its completion condition is satisfied" (soon).]

So currently the word "complete" is a bit overloaded, but at least complete and incomplete tasks  are logical opposites. And that's important because it describes when the scheduler can forget (as complete) or retain (as incomplete) a finished task.

If we change to *a task is complete if its output completion condition is satisfied* (even if it's not finished), then logically a task should be incomplete if its output completion condition is NOT satisfied (even if not finished). But we still have *a task is incomplete, if the completion condition is not satisfied but the task has a final task status*.

I think we need to change the definition of incomplete too:
- a task is complete if its output completion condition is satisfied
- a task is incomplete if its output completion condition is not satisfied
- a finished task will be removed if complete, and retained if incomplete

hjoliver: Following up on that, let's go with your use of "satisfied" instead of "completed" for individual outputs as well. We've always said that prerequisites/dependencies get satisfied and outputs get completed, which makes sense, but in the new world perhaps its OK to say that "task outputs get satisfied by task messages (or by `cylc set`). That will remove all ambiguity from the concept of "completion" in Cylc:

- individual outputs get satisfied (not completed) by task messages or `cylc set`
- a task is complete if its output completion condition is satisfied, otherwise it is incomplete
- finished tasks get removed from the active pool if complete, or retained if incomplete

OK?

hjoliver: (I think that's my only problem with the amended amended proposal).

oliver-sanders: > a task is incomplete if its outputs are not complete and it is finished

Read down the proposal a couple lines, this is already in there.

oliver-sanders: > let's go with your use of "satisfied" instead of "completed" for individual outputs

I'm not fussed by grammar gerrymandering. I would lean towards:

* An expression can be "satisfied".
* A task output can be "completed".

If you want something different let me know.

oliver-sanders: > I think that's my only problem with the amended amended proposal, otherwise great

I think the issue is grammar? Assuming there's no fundamental issue with this proposal, I'm going to push ahead with the branch in the name of haste.

hjoliver: > >    a task is incomplete if its outputs are not complete and it is finished

> Read down the proposal a couple lines, this is already in there.

I know. My point is: that is now logically inconsistent with your new definition of "completed task".

> Sorry, I'm a bit confused about what you're suggesting. This proposal states:
> >    A task is complete if the completion condition is satisfied.
>>  And:
> >   a task is incomplete, if the completion condition is not satisfied
    but the task has a final task status (note status not output). This is the
    same as presently implemented (8.2.x)
>
> Ok?

I agree (well, I observe) that the proposal states both of those things, but the problem is that means "complete" and "incomplete" tasks are no longer logically opposite - only one of them cares about final task status - and I think that is a problem for clarity and user understanding.

> >  let's go with your use of "satisfied" instead of "completed" for individual outputs
> I'm not fussed by grammar gerrymandering. I would lean towards:
>    - An expression can be "satisfied".
>    - A task output can be "completed".
> If you want something different let me know.

This isn't grammar, it's about the definition of the words that we're using to describe these concepts.  In this case, I also lean towards those existing uses, but I thought **you** wanted something different! I was literally quoting **your wording** in the proposal ("satisfying" rather than "completing" an output), assuming that was deliberate, and speculating that your rationale was to disambiguate our current overloaded use of the word "complete".

> I think the issue is grammar? Assuming there's no fundamental issue with this proposal, I'm going to push ahead with the branch in the name of haste. Let me know if I've misinterpreted this.

It's not grammar, as I said, it's about defining terms clearly and consistently. But yes, that's still not a fundamental issue, so please do go ahead with implementation (sorry I was under the impression you already were working on it, and didn't realize you might think this was a "fundamental" objection).

hjoliver:
@oliver-sanders - I'll try to distill my two "definitional" issues further. This is important because we all have to agree on what the terms mean, and document the system accordingly.

### complete vs incomplete tasks

Task completion should depend ONLY on outputs, not on task state [original comment](https://github.com/cylc/cylc-admin/pull/189#discussion_r1529343688)
  - a task is complete if its output completion expression is satisfied
  - a task is incomplete if its output completion expression is not satisfied  `#<--- NEW (no final status involved)`
  - removal from the active task pool depends on completion AND having a final status (a.k.a. "finished")

(Without this, as proposed complete and incomplete tasks are not logical opposites - you can have a task that is both not complete and not incomplete 🤯 )

### satisfying vs completing outputs

Individual outputs should be "satisfied", not "completed" [original comment](https://github.com/cylc/cylc-admin/pull/189#discussion_r1529363776) 
   - this would disambiguate our overloaded use of the term "complete"
   - it can be spun as making good sense too: a task message "satisfies" a task output
  
(Note, I thought, from your wording that **you** were proposing this; I was just trying to take it seriously 😁. Either way, it is less important than 1., but worth considering)

oliver-sanders:
I'm ok with complete and incomplete not being literal opposites, like flammable and inflammable, hence it's a grammar argument, happy to accept suggestions, but I don't have any myself.

Happy to fiddle the meaning of completed and satisfied, but I don't think I'm going to be changing terminology universally in this proposal/PR as it's more far reaching than `task_outputs.py`.

dpmatthews:
We're going to need to support an event (and indicate it in the GUI) if a task can not complete so we need to be able to describe this clearly. I agree that `incomplete` would not be a good name for this event. Maybe `wont_complete`?

hjoliver:
This isn't one of those "legendary" arguments (the one you referenced was a semi-humorous suggestion for a collective noun where one didn't exist). We have to have an agreed meaning of important terms like this or we can't understand each other or expect users to. And it's not grammar (i.e., it not abou the structure of language and sentences) it's just about consistent and logical definitions of specific terms.

But OK, since you don't seem to have strong feelings on the matter, except that you think it's grammar and I don't, how about we go with my proposed definitions which are both intuitive and logically consistent:
- a task is complete if it's completion condition is satisfied 
- a task is incomplete (i.e. not complete) if it's completion is not satisfied
- (and therefore a task can be removed from the pool if it is complete AND finished (i.e., it has a final status)

hjoliver:
My suggestion also handles @dpmatthews point - the event that we need to signal is that a *task finished without completing*. "Incomplete" is not a good name for that. "wont_complete" is a bit ungainly, but it makes sense because a finished task can no longer complete automatically. Maybe we can think of a better name for that.

oliver-sanders:
> a task is incomplete (i.e. not complete) if it's completion is not satisfied

I don't think your suggestion works? E.G. if a task has no required outputs, then it will be completed before it has even been spawned.

This is what I'm proposing, the task status part is required:

* Outputs are complete if the completion expression is satisfied.
* Tasks are complete if they have a final task status AND their outputs are complete.

hjoliver:
> I don't think your suggestion works? E.G. if a task has no required outputs, then it will be completed before it has even been spawned.

That works fine, if (as I'm proposing) "completion" refers only to completion of outputs. If such a task gets triggered, it is not required to generate any outputs at all, it just needs to finish.  

> This is what I'm proposing:
>   - Outputs are complete if the completion expression is satisfied.
>   - Tasks are complete if they have a final task status AND their outputs are complete.

That's not quite what you are proposing. The proposal refers to both complete and incomplete tasks:
> - *A task is incomplete if the completion condition is satisfied and the task has a final status.*

Those words are logically opposite: something is either complete or incomplete, end of story. But your use of these terms does not conform that: **as proposed, a task that is "not complete" is not necessarily "incomplete".** That is not consistent (aka inconsistent!) with the normal definition of the words, and it will cause confusion.

> This is not inconsistent, it is the required logic. ... 

The required code logic is 100% the same either way, but language used in the proposal to describe it is inconsistent with the universally agreed meaning the these terms. And that, almost by definition, is confusing.

> Note these definitions control internal logic and aren't user facing, we report this to the users as "task finished with incomplete outputs".

Our proposals do necessarily also invent and use terminology to describe what's going on, and which then has to be documented in the user guide.

To avoid this inconsistency in the descriptive terminology, which is both developer- and user-facing, we have two choices:

1.  a task is described as "complete" or "incomplete" based on only output completion
     - task pool removal logic depends on both "task completion" and "task finished"
2. we don't use the terms "complete task" and "incomplete task" at all, only *outputs* can be complete or incomplete
     - task pool removal logic depends on both "task output completion" and "task finished"
     
These are both the same implementation-wise, and they both avoid the inconsistency that I'm complaining about. But I prefer 1. It seems perfectly fine to me to say that a task is complete if its required outputs are complete, but that it has to be finished as well before we are done with it.

hjoliver:
> Feel free to suggest different names if you feel strongly.

Ok, I think 2. above is my suggestion. Reasons:
- we can't document this new logic with use of complete/incomplete as described above
- but I accept your point that users might be confused that a task can be complete without finishing

So let's just not say that a *task* can be in/complete.  It's outputs can be complete or not, and it can be finished or not.

docs/proposal-optional-output-extension.md
```diff
@@ -41,16 +41,33 @@ There are currently three problems with expiry (as of cylc-flow 8.1.4):
    executed and that might not happen if expiry or submission are optional (or if the pre-requisites
    for `a` are not met).
 
-3. A task is incomplete if the completion condition is not satisfied and:
+3. A task is complete if the completion condition is satisfied.
```
oliver-sanders, line 44:
```suggestion
3. Outputs are complete if the completion condition is satisfied.
```

oliver-sanders: @hjoliver is this correction sufficient for you?

hjoliver: Not by itself - it has to be carried throughout the document wherever "complete tasks" and "incomplete tasks" are referred to.

I take it you are agreeing to [2. above](https://github.com/cylc/cylc-admin/pull/189#issuecomment-2010574702) then?

hjoliver: Happy to have a crack at that myself if you're busy on implementation?

oliver-sanders: > I take it you are agreeing to https://github.com/cylc/cylc-admin/pull/189#issuecomment-2010574702 then?

No, I'm sticking with:

* Outputs are complete if the completion expression is satisfied.
* Tasks are complete if they have a final task status AND their outputs are complete.

Reminder, `TaskOutputs.is_complete` is an internal interface not exposed to users, this definition is not important.

Providing there aren't any technical qualms with this, I'm going to plug on with implementation as there's still quite a lot to do.

hjoliver: But are you also dropping your proposed definition of *incomplete tasks*? If not, we're at an impasse. 

My bottom line is we can't redefine the terms complete and incomplete so that `"not complete" != "incomplete"`

Yes this isn't technical in the sense that we agree entirely on the logic, but it sure would be nice if the implementation didn't use inconsistent terms too.

hjoliver: > eh? Tasks and TaskOutputs are two different things. I see no issue with what is written.

Your single-line change does not address my concerns without additional information.  As far as I can tell you are now saying that you won't call tasks complete if their outputs are complete, but you will still call them incomplete if their outputs are incomplete AND they are finished. That's a bit of a mess compared to either of my two ways to fix the inconsistent terminology.

You could stop spending time on this by simply agreeing - this is a review of a proposal and I've raised a valid objection to the terminology being used.

oliver-sanders: As far as I can tell you are now saying that you won't call tasks complete if their outputs are complete, but you will still call them incomplete if their outputs are incomplete AND they are finished

Correct.

> You could stop spending time on this by simply agreeing

Harsh. Your turn to propose something, my turn to be nit picky.

hjoliver: Sorry, didn't mean to be harsh - but I have proposed a solution (two of them in fact), in great detail, above. This is *only* a small question of consistent use of two common terms, it doesn't require a whole new proposal. Are you sure you've read and understood my "proposals"? (This sort of thing would be a lot easier to resolve if we didn't live in opposite timezones!)

hjoliver: How about I merge this and immediately put up a tweak PR that redoes the use of "in/complete tasks" in the way that I propose?  (However, don't take that as full approval of this one yet :grin:)

oliver-sanders: If you could pick an optional and raise it so I can review it in context.

hjoliver: Ok, I'll clone your branch (rather than merge it) and put up an alternative.

PR #190, opened 2024-03-22
amendment to amendent of Proposal amendment 

hjoliver:
My small tweak of #189 - terminology only, but it's critical terminology for how we document this functionality.  

My (only!) problem with #189 is that it changes the meaning of terms in such a way that `incomplete task` no longer means the same thing as `a task that is not complete`.  Given the universally agreed meaning of those very commonly used words, we can't do that.

We have to (or at least it makes undeniable sense to) describe outputs as "complete" or "incomplete", but we do not have to call tasks "complete" or "incomplete". The only essential concepts are:
- is the task "finished" or not (i.e., does it have a final task status)
- are its outputs complete or not

So, my proposed amendment to the proposed amendment to the amended proposal is essentially this:
- task outputs can be described as complete or incomplete
- tasks can be described as finished or not finished (final task status)
- task pool removal depends on tasks being finished with complete outputs

As non-essential bonus I propose we shorten (when appropriate) "finished with complete outputs" to "done" (the task has done its job in the workflow and the scheduler can forget it)

This branch is a clone of 189, with another commit (mine) on top. So the full diff includes all the changes of @oliver-sanders plus mine. Look at the final commit alone to see my on-top changes.

### DIFF
--- a/docs/proposal-optional-output-extension.md
+++ b/docs/proposal-optional-output-extension.md
@@ -41,16 +41,33 @@ There are currently three problems with expiry (as of cylc-flow 8.1.4):
    executed and that might not happen if expiry or submission are optional (or if the pre-requisites
    for `a` are not met).
 
-3. A task is incomplete if the completion condition is not satisfied and:
+3. A task is complete if the completion condition is satisfied.
 
-   * It finished executing
+   The completion condition can be defined by the user, otherwise the
+   expression is automatically generated according to the rules:
 
-   * OR Job submission failed and the `:submit` output was not optional
+   1. All required outputs (referenced in the graph) must be satisfied.
+   2. If success is optional, then the task is complete if it fails.
+   3. If submission is optional, then the task is complete if it submit-fails.
+   4. If expiry is optional, then the task is complete if it expires.
 
-   * OR The task expired and the `:expire` output was not optional
-     * This is a new condition to handle expiry
+   I.E:
+
+   ```python
+   is_complete = (
+     all(output for output in outputs if not is_optional(output))
+     or (failed if is_optional(succeeded))
+     or (submit_failed if is_optional(submitted))
+     or (expired if is_optional(expired))
+   )
+   ```
+
+   Note, a task is *in*complete, if the completion condition is not satisfied
+   but the task has a final task status (note status not output). This is the
+   same as presently implemented (8.2.x)
 
 4. `:expire` cannot be required.
+
    Similarly `:submit-fail` cannot be required.
    
    * We currently allow `:submit-fail` to be required but this doesn't really make sense.
@@ -69,6 +86,8 @@ There are currently three problems with expiry (as of cylc-flow 8.1.4):
    # OK
    completion = succeeded or failed
    completion = succeeded and (x or y or z)
+   completion = (succeeded and x) or (failed and y)
+   completion = (succeeded and (x or y or z)) or failed or expired
 
    # ERROR
    completion = not failed
@@ -78,8 +97,8 @@ There are currently three problems with expiry (as of cylc-flow 8.1.4):
 
    Expression to be validated (i.e. parsed) at validate time.
 
-   If the completion expression is defined, any outputs which are optional in the expression
-   should be marked as optional in the graph if used in the graph:
+   If the completion expression is user-defined, then it must be logically
+   consistent with the graph.
 
    ```
    completion(a) = succeeded and (x or y or z)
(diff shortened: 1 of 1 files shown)


docs/proposal-optional-output-extension.md
```diff
@@ -62,14 +62,36 @@ There are currently three problems with expiry (as of cylc-flow 8.1.4):
    )
    ```
 
-   Note, a task is *in*complete, if the completion condition is not satisfied
-   but the task has a final task status (note status not output). This is the
-   same as presently implemented (8.2.x)
+   Similarly, a task's outputs are "incomplete" or "not complete" if the completion
+   condition is not satisfied.
+
+3.5 A task is "finished" if it has final task status
```
dpmatthews, line 68: "Finished" already has a well established meaning (succeeded or failed) so I don't think we should use that. I suggest we stick with "final status" (as you've done in 3.6 below) or "final task status".

docs/proposal-optional-output-extension.md
```diff
@@ -41,18 +41,57 @@ There are currently three problems with expiry (as of cylc-flow 8.1.4):
...
+
+   A finished task with complete outputs can be removed as "done".
+
+   A finished task with incomplete outputs is retained as "not done",
+   pending intervention to complete its outputs or to remove it manually.
+
+  N.B. giving a short name for "finished with complete outputs" is not strictly
+  necessary, but it is convenient. I propose "done": a that finishes with completed
+  outputs has done its job and can be forgotten by the scheduler. "Not done" is the
+  logical opposite of that.
```
dpmatthews, line 88: Now that we are distinguishing between a task and it's outputs I don't don't think we need a new name.

A tasks output's are either "complete" or "incomplete" (I think we're all happy with that).

A task can be:

- not final
- final status with complete outputs
  - I think it's fine to refer to this as a "complete" task (we're talking about a task here, not it's outputs)
- final status with incomplete outputs
  - I think it's fine to refer to this as an "incomplete" task
  - As long as we define an "incomplete" task as "a task that has reached final status with incomplete outputs" I don't see any cause for confusion
  - (I think it would too easy to confuse "not done" with "not run" or "not finished")

docs/proposal-optional-output-extension.md
hjoliver: Well, succeed and failed are the two primary final statuses.  But I agree the old `:finished` pseudo output is a spanner in those works.

So ok, "final status" it is.

docs/proposal-optional-output-extension.md
hjoliver: 🎉 Haha, that was an unholy amount of discussion just to get back to exactly what we have now.

My entire objection was the proposed change to *A task is complete if the completion condition is satisfied.* instead of *final status and complete outputs*, which results in *incomplete != not complete*.

Yeah we certainly don't need "done" if we continue to use "complete task" in this way, which is exactly what I meant by "done".

We don't particularly need to call tasks "complete" or "incomplete", but it's a reasonable overloading of the two terms so long their definitions are self-consistent.

I'll make final tweaks to the, then we're "done".

hjoliver: No, a task can be:
- not final and incomplete outputs
- not final and complete outputs
- final and complete outputs
- final and incomplete outputs

Your proposed definition of a "complete task" is:
- final status and complete outputs
- *(so far so good)*

Therefore an "incomplete task" (i.e., "not complete") is:
- not (final and complete outputs), i.e., **not final or incomplete outputs**
- *(this includes waiting tasks, and active tasks that have not completed their outputs - and it is perfectly sensible to call those "incomplete")*

But you evidently want to define (your words) an "incomplete task" as:
- final status and incomplete outputs

That contradicts the basic "logic of completeness" above.

It's completely unnecessary to do that. We just need to say:
- a complete task is final and has complete outputs
- otherwise the task is incomplete
- task pool logic is:
  - remove tasks that are final and complete
  - retain tasks that are final and incomplete

oliver-sanders: > I'm confused. A task can be:
>
>> No, a task can be:

@hjoliver, you appear to have agreed with Dave here rather than contradicting him?

These are the cases (you wrote above):

* not final and incomplete outputs
* not final and complete outputs
* final and incomplete outputs
* final and complete outputs

There is no need to differentiate between the non-final cases, so these are the cases we need to consider (as Dave wrote above):

1. not final
2. final and incomplete outputs
3. final and complete outputs

We don't need a name for (1), this is covered by other terminology, but we do need names for (2) and (3). I had kept "complete" and "incomplete" for these. You have suggested "done" and "not done".

> Now that we are distinguishing between a task and it's outputs I don't don't think we need a new name.

Since task completion (user-facing) and output completion are separate things, renaming "complete" to "done" doesn't change anything by itself. As such, I don't see the need to introduce a new term here and redefine user facing concepts in terms of it. So we could just rename "done" back to "complete" in your new definition, this wont change its meaning.

dpmatthews: We currently define an "incomplete" task as a task that finishes without completing all required outputs.
https://cylc.github.io/cylc-doc/stable/html/glossary.html#term-incomplete-task
We now need to tweak this to say final status rather than finishes but fundamentally I'm proposing we keep the existing definition which is widely used in the documentation. I don't see any logical inconsistently - a task can can only be complete or incomplete once it has reached a final status.

Note that "final with incomplete outputs" is an important state that we need an easy way to reference. We need to flag tasks in this state in the GUI and add an event for this state (https://github.com/cylc/cylc-flow/issues/4957).

hjoliver: OK ... here's my response.

Using "complete" and "incomplete" in this way IS inconsistent with the normal meaning of those terms, despite both your claims to the contrary. They are not instantaneous binaries like "succeeded" and "failed".  Just Google the meaning of incomplete" - anything that becomes complete is logically incomplete *until* that point. It doesn't suddenly become complete or incomplete in an instant. 

@dpmatthews you said (on Element) *It's no more inconsistent than it used to be!* ... which is clearly an admission that it is inconsistent!  

Further, it's completely unnecessary to do this. We could easily (as per my unassailable logic above) just say that tasks achieve a final status with complete or incomplete outputs, and act accordingly.

IMO this is likely to confuse users, because they will naturally assume that they know what complete and incomplete mean, especially when we also link the concept to completion of outputs. 

How much of a problem that is, is at this point a matter of opinion, but I'm not ecstatic about voluntarily choosing inconsistent terminology when we really don't have to do it.

The ONLY point in favour of this is our own recent precedent: we already use the "incomplete task" in this way. But:
- adding the definition of "complete task" in opposition to this makes the inconsistency more glaring
- and do you think many users will be put out by fixing this now? I'm pretty sure no one at NIWA is invested in the concept yet - we've been holding off pushing this stuff out till 8.3.0.

BUT if you guys really ain't gonna back down on this, I can live with it so long as we explicit document that "complete task" and "incomplete task" have this weird meaning - whereby a task is not incomplete prior to achieving completion. Which does frankly warrant another use of the 🤯 emoji).

oliver-sanders: > @dpmatthews you said (on Element) It's no more inconsistent than it used to be! ... which is clearly an admission that it is inconsistent!

Dave is pointing out that the "inconsistency" you perceive was not introduced by my amendment, but comes from the original proposal.

> Further, it's completely unnecessary to do this. We could easily (as per my unassailable logic above) just say that tasks achieve a final status with complete or incomplete outputs, and act accordingly.

Ok, it's going to be easier to bypass this argument than resolve it. I had already adopted your wording on https://github.com/cylc/cylc-admin/pull/191 which you may have missed. Please see if this works for you.

oliver-sanders (changes requested):
Would prefer not to introduce "done" and "not done" as using facing concepts.

Note also `done !== not done` :grin:

Hopefully https://github.com/cylc/cylc-admin/pull/191 will suffice?

docs/proposal-optional-output-extension.md
hjoliver: > Dave is pointing out that the "inconsistency" you perceive was not introduced by my amendment, but comes from the original proposal.

Yes, I acknowledged that (and thus take some responsibility for the situation myself). But the amendment did make the inconsistency more obvious, and we can and should still fix things we got wrong in the past.

> Ok, it's going to be easier to bypass this argument than resolve it

I think it's interesting that you both keep disagreeing with me on this without ever actually countering my argument other than to say you don't see any inconsistency (which I don't believe, since `incomplete != not complete` is transparently inconsistent!).   

But ok ...

> I had already adopted your wording on https://github.com/cylc/cylc-admin/pull/191 which you may have missed. Please see if this works for you.

... I'll take a look - I had missed it sorry, in a bit of a rush after returning from Easter break.

hjoliver:
> Would prefer not to introduce "done" and "not done" as using facing concepts.

That was just my attempt to introduce self-consistent terms for when a task can be removed from the pool or not, by divorcing the terminology from the concept of output completion.

> Note also `done !== not done` 😁

Is that a typo?  I certainly agree that `done != not done`!  The problem here is that `incomplete != not complete`.  My tentative "done" proposal was that a task is "done" when final with complete outputs, and not done otherwise, which is self-consistent.

> Hopefully #191 will suffice?

Hopefully! ...

hjoliver:
Yep, all good, phew!

PR #191, opened 2024-03-28
amendment to proposal amendment

oliver-sanders:
Following on from our conversation the other day, here is a new amendment that I hope will satisfy all:

    
* Redefine task output "incompletion" to avoid problem edge cases.
* Permit pre-conditions in completion expressions.
* Remove redundent safety check that was only required for the previous proposal
* Remove references to task completion/incompletion in line with Hillary's preferences.


hjoliver (approved):
Happy now, thank 🎉

PR #192, opened 2024-04-10
release: add note to announce new features on discourse

oliver-sanders:
Something I've been doing for the last couple of releases that I think is worth continuing with.


hjoliver (approved):
Yep, nice.

PR #213, opened 2025-09-24
Create repository-dependencies.md

wxtim:
I created this for @samuel-denton, and suggested that it was added to metomi-docs, then realized that it was probably best stored here, since it's a more generally useful diagram.


oliver-sanders:
This can go into the developer docs, suggest adding to this page: https://cylc.github.io/cylc-admin/howto/dev-setup

PR #216, opened 2025-10-02
Update Meta-release data

wxtim:


oliver-sanders (approved):
(best wait for all the branches to be created and the intra-component dependencies to be updated before merging)

PR #228, opened 2026-05-14
fix url

oliver-sanders:
Sorry, jiggered up the URL of this image.

Don't think it can be tested other than by inspection / being hosted.


samuel-denton (approved):
LGTM

